PR #3, opened 2019-01-18
Add maintainers and CODEOWNERS

vaporos:


vaporos:
Sorted both contents of maintainers and codeowners files alphabetically.

PR #13, opened 2019-03-19
Add initial Grid Schema Protos

agunde406:
This required temporarily renaming Track and Trace's PropertyValue to TrackAndTracePropertyValue because protobuf would not build two instances of PropertyValue. We plan to update Track And Trace to use Grid Schema once it has been completed.

### DIFF
--- a/sdk/protos/schema_payload.proto
+++ b/sdk/protos/schema_payload.proto
@@ -1,0 +1,55 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


sdk/protos/schema_payload.proto
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+syntax = "proto3";
+
+import "schema_state.proto";
+
+message SchemaPayload {
```
peterschwarz: Add some doc comments to these.

PR #14, opened 2019-03-19
Add initial "grid" CLI

vaporos:


cli/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
agunde406: missing license

cli/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+// limitations under the License.
+
+use std::error::Error;
+use std::fmt;
+
+use log;
+
+#[derive(Debug)]
+pub enum CliError {
+    LoggerError(Box<log::SetLoggerError>),
```
peterschwarz: This `SetLoggerError` only happens if the `init_with_level` has been called more than once.  The user can't do anything about that, and so it's fundamentally a programming error.

See [SetLoggerError](https://docs.rs/log/0.4.6/log/struct.SetLoggerError.html), which is ultimately used [here](https://github.com/borntyping/rust-simple_logger/blob/master/src/lib.rs#L49) in simple logger.

vaporos: Are you advocating that any programming error should cause an immediate panic, or that the message output should be better?

peterschwarz: I'm advocating a panic in this particular case.  If the user were to receive this, they would always receive it.  I.e. they can't fix this error through any combination of flags.  Since this sort of error would be caught at programming time (for the same reason) a panic makes more sense than adding an value to the enum.

peterschwarz: For other programming errors, I would consider evaluating it on a case-by-case basis.

cli/Cargo.toml
vaporos: Fixed

cli/src/error.rs
vaporos: I get your point, but I think we shoudn't fall into the pattern of scattering panics throughout the code. In this case, it is easy to avoid the panic. In the case of locking-related panics, I think we are using excepts because propagating up the unlikely errors have a big impact on the code overall, and so its a trade-off. There is no trade-off here though - the code looks cleaner "?" and is more technically correct (prints error in main() and exits with a status code).

After logging is setup, we should actively avoid panics because they will not get logged appropriately. Obviously, not justification for this particular case. But if in every other instance we are avoiding panics so that errors get logged appropriately, then I don't think we should deviate here, we should just follow the more correct pattern of turning errors (whether they be user-caused or not) into CliErrors and let it get handled appropriately in main().

vaporos: I updated this from LoggingError to LoggingInitializationError to make it clearer.

PR #16, opened 2019-03-21
Add initial daemon package

shannynalayna:
Adds initial daemon package within Grid and adds the generated build docs from this package to the .gitignore file. Currently, none of the contents of this package are really functional.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 Bitwise IO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


daemon/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+
+fn main() {
```
vaporos: We could put some boilerplate code here from the CLI command re: argument parsing, logging setup, initial top-level error handling.

shannynalayna: Addressed

PR #19, opened 2019-03-22
Add gridd to docker compose file

peterschwarz:
Add a Dockerfile for `gridd` and include a container in the docker compose file.


daemon/Dockerfile
```diff
@@ -0,0 +1,69 @@
+# Copyright 2018 Cargill Incorporated
```
agunde406: 2019

PR #20, opened 2019-03-25
Add initial REST API specs

eloaverona:
This PR adds initial openapi spec doc for the Grid REST API. 
It defines endpoint to submit batches, get batch status, and get schema from the reporting database. 
To visualize the document [Swagger](http://editor.swagger.io/) is a good tool.

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -109,6 +109,76 @@ paths:
           $ref: "#/responses/500ServerError"
         503:
           $ref: "#/responses/503ServiceUnavailable"
+
+  /schema:
+    get:
+      tags:
+      - "schema"
+      summary: "Get a list of schemas"
+      description: "Fetches a list of schemas from the reporting database"
+      operationId: "get_schemas"
+      produces:
+      - "application/json"
+      responses:
+        202:
+          description: "List of schemas"
+          schema:
+            properties:
+              data:
+                type: array
+                items:
+                 $ref: "#/definitions/Schema"
+
+        400:
+          $ref: "#/responses/400BadRequest"
+        429:
+          $ref: "#/responses/429TooManyRequests"
+        500:
+          description: Something went wrong within the database
+          schema:
+            $ref: "#/definitions/Error"
+        503:
+          description: API is unable to reach the database
+          schema:
+            $ref: "#/definitions/Error"
+
+  /schema/{schema_name}:
+    get:
+      tags:
+      - "schema"
+      summary: "Find schema by schema name"
+      description: "Returns a single schema"
+      operationId: "get_schema_by_name"
+      produces:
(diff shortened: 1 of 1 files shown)


daemon/openapi.yaml
```diff
@@ -0,0 +1,393 @@
...
+# ------------------------------------------------------------------------------
+
+swagger: '2.0'
+
+info:
+  version: "0.1.0"
+  title: Grid REST API
+  description:
+    _An API providing HTTP/JSON interface to Hyperledger Grid simplifying client
+    interaction with the blockchain by exposing endpoints that use common
```
agunde406: Since Grid could be used without a blockchain, we should remove the mention here.

daemon/openapi.yaml
```diff
@@ -0,0 +1,393 @@
...
+    HTTP/JSON standards._
+
+paths:
+  /batches:
+    post:
+      tags:
+      - sawtooth_validator
+      summary: Sends a BatchList to the validator
```
agunde406: Mention Sawtooth Validator instead of just validator

daemon/openapi.yaml
```diff
@@ -0,0 +1,393 @@
...
+  wait:
+    name: wait
+    in: query
+    type: integer
+    description: A time in seconds to wait for commit
+
+definitions:
+  Link:
+    type: string
+    example: https://api.sawtooth.com/state?head=65cd3a3ce088b265b626f704b7f3db97b6f12e848dccb35d7806f3d0324c71b709ed360d602b8b658b94695374717e3bdb4b76f77886953777d5d008558247dd
```
agunde406: replace https://api.sawtooth.com with https://api.grid.com

daemon/openapi.yaml
```diff
@@ -0,0 +1,393 @@
...
+    properties:
+      name:
+        type: string
+        example: "Lightbulb"
+      description:
+        type: string
+        example: "Example Lightbulb schema"
+      owener:
+        type: string
+        example: "philips"
```
agunde406: if we want to match the rfc this could be philips001

daemon/openapi.yaml
eloaverona: Addressed

daemon/openapi.yaml
eloaverona: Addressed

daemon/openapi.yaml
eloaverona: Addressed

daemon/openapi.yaml
eloaverona: Addressed

vaporos:
Please add a note in the commit messages which describe the documents these are derived from, for reference purposes.

PR #28, opened 2019-03-29
Initial Event Processor

peterschwarz:
This event processor is configured to receive the appropriate events from a Sawtooth Validator, which can be offloaded into a reporting db (to be added in a subsequent PR).


daemon/src/event/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+        let event_handler = event_handlers
+            .iter()
+            .find(|handler| handler.event_type() == event.get_event_type());
+
+        if let Some(event_handler) = event_handler {
+            if let Err(err) = event_handler.handle_event(&event) {
+                error!("Unable to handle event {}: {}", event.get_event_type(), err);
+            }
+        } else {
+            warn!("Unable to handl event {}", event.get_event_type());
```
jsmitchell: s/handl/handle/

daemon/src/main.rs
```diff
@@ -53,18 +58,45 @@ fn run() -> Result<(), DaemonError> {
...
+    )
+    .map_err(|err| DaemonError::EventProcessorError(Box::new(err)))?;
+
+    let (event_processor_shutdown_handle, event_processor_join_handle) =
+        evt_processor.take_shutdown_controls();
 
+    let ctrlc_triggered = AtomicBool::new(false);
     ctrlc::set_handler(move || {
+        if ctrlc_triggered.load(Ordering::SeqCst) {
+            process::exit(0);
```
vaporos: This should probably be process::exit(1). This should maybe print something out to stderr prior to exiting like "Aborting due to multiple Ctrl-C events...".

daemon/src/rest_api/mod.rs
```diff
@@ -55,21 +54,31 @@ pub fn run(
...
+                .start();
 
             tx.send(addr).map_err(|err| {
                 RestApiError::StartUpError(format!("Unable to send Server Addr: {}", err))
             })?;
 
-            Ok(sys.run())
+            sys.run();
+
+            info!("Rest API terminiating");
```
vaporos: Typo

docker-compose.yaml
```diff
@@ -97,11 +97,21 @@ services:
...
       context: .
       dockerfile: daemon/Dockerfile
     expose:
       - 8080
     ports:
       - '8080:8080'
-    entrypoint: "/gridd -v -b gridd:8080"
+    entrypoint: /gridd -vv -b gridd:8080 -C tcp://validator:4004
+
+  sawtooth-shell:
```
vaporos: We are probably going to want to create our own grid shell instead and install sawtooth and grid CLIs into it.

peterschwarz: I'll leave that to a future PR.

PR #30, opened 2019-04-02
Submit batches endpoint

eloaverona:
Implements POST submit_batches endpoint as specified by the openapi spec documentation.

### DIFF
--- a/daemon/src/rest_api/error.rs
+++ b/daemon/src/rest_api/error.rs
@@ -12,35 +12,117 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use actix::MailboxError;
+use actix_web::error::{PayloadError, ResponseError, UrlGenerationError};
+use actix_web::HttpResponse;
 use std::error::Error;
+
 use std::fmt;
 
 #[derive(Debug)]
-pub enum RestApiError {
+pub enum RestApiServerError {
     StartUpError(String),
     StdError(std::io::Error),
 }
 
-impl Error for RestApiError {
+impl From<std::io::Error> for RestApiServerError {
+    fn from(err: std::io::Error) -> RestApiServerError {
+        RestApiServerError::StdError(err)
+    }
+}
+
+impl Error for RestApiServerError {
     fn source(&self) -> Option<&(dyn Error + 'static)> {
         match self {
-            RestApiError::StartUpError(_) => None,
-            RestApiError::StdError(err) => Some(err),
+            RestApiServerError::StartUpError(_) => None,
+            RestApiServerError::StdError(err) => Some(err),
         }
     }
 }
 
-impl fmt::Display for RestApiError {
+impl fmt::Display for RestApiServerError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            RestApiError::StartUpError(e) => write!(f, "Start-up Error: {}", e),
-            RestApiError::StdError(e) => write!(f, "Std Error: {}", e),
+            RestApiServerError::StartUpError(e) => write!(f, "Start-up Error: {}", e),
+            RestApiServerError::StdError(e) => write!(f, "Std Error: {}", e),
         }
     }
 }
 
-impl From<std::io::Error> for RestApiError {
-    fn from(err: std::io::Error) -> RestApiError {
-        RestApiError::StdError(err)
+#[derive(Debug)]
+pub enum RestApiResponseError {
+    BadRequest(String),
+    SawtoothConnectionError(String),
+    SawtoothValidatorResponseError(String),
+    RequestHandlerError(String),
+}
+
+impl Error for RestApiResponseError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        match self {
+            RestApiResponseError::BadRequest(_) => None,
+            RestApiResponseError::SawtoothConnectionError(_) => None,
+            RestApiResponseError::SawtoothValidatorResponseError(_) => None,
+            RestApiResponseError::RequestHandlerError(_) => None,
+        }
+    }
+}
+
+impl fmt::Display for RestApiResponseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            RestApiResponseError::BadRequest(ref s) => write!(f, "Bad Request: {}", s),
+            RestApiResponseError::SawtoothConnectionError(ref s) => {
+                write!(f, "Zmq Connection Error: {}", s)
+            }
+            RestApiResponseError::SawtoothValidatorResponseError(ref s) => {
+                write!(f, "Sawtooth Validator Response Error: {}", s)
+            }
+            RestApiResponseError::RequestHandlerError(ref s) => {
+                write!(f, "Sawtooth Validator Response Error: {}", s)
+            }
+        }
+    }
+}
+
+// impl ResponseError trait allows to convert our errors into http responses with appropriate data
+impl ResponseError for RestApiResponseError {
+    fn error_response(&self) -> HttpResponse {
+        match *self {
+            RestApiResponseError::BadRequest(ref message) => {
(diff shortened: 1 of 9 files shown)


daemon/src/rest_api/error.rs
```diff
@@ -12,20 +12,32 @@
...
+use actix_web::{error::ResponseError, HttpResponse};
 use std::error::Error;
+
 use std::fmt;
 
 #[derive(Debug)]
 pub enum RestApiError {
     StartUpError(String),
     StdError(std::io::Error),
+    InternalServerError,
```
peterschwarz: We should consider splitting this error up into something that is for the overall rest api component (i.e. starting, stopping, etc) and handling requests.

daemon/src/rest_api/route_handler.rs
```diff
@@ -12,8 +12,158 @@
...
+        }
+    }
+}
+
+pub fn submit_batches(
+    (req, state): (HttpRequest<AppState>, State<AppState>),
+) -> impl Future<Item = HttpResponse, Error = RestApiError> {
+    req.body().from_err().and_then(
+        move |body| -> Box<Future<Item = HttpResponse, Error = RestApiError>> {
+            let batch_list: Result<BatchList, _> = protobuf::parse_from_bytes(&*body);
```
peterschwarz: The parse error should be handled here, not below:
```
let batch_list: BatchList = match protobuf::parse_from_bytes(&*body) {
    Ok(list) => list,
    Err(err) => return Box::new(future::err(RestApiError::BadRequest(format!(
                    "Protobuf message was badly formatted. {}",
                    err.to_string()
                ))));
};
```
This keeps all the error result creation close to the originator of the error.

daemon/src/rest_api/route_handler.rs
```diff
@@ -12,8 +12,158 @@
...
+                    let res = state
+                        .zmq_connection
+                        .send(SubmitBatches {
+                            batch_list,
+                            response_url,
+                        })
+                        .from_err()
+                        .and_then(|res| match res {
+                            Ok(link) => Ok(HttpResponse::Ok().json(link)),
+                            Err(_) => Ok(HttpResponse::InternalServerError().into()),
```
peterschwarz: Some of the errors may imply poorly formed batches, which can be communicated back to the user.

daemon/src/rest_api/route_handler.rs
```diff
@@ -12,8 +12,158 @@
...
+    _receiver: MessageReceiver,
+}
+
+impl Actor for ZmqConnection {
+    type Context = Context<Self>;
+}
+
+impl ZmqConnection {
+    pub fn new(validator_address: String) -> ZmqConnection {
+        let zmq_connection = ZmqMessageConnection::new(&validator_address);
```
peterschwarz: How will this be closed?  This object can probably make due with just the sender, and we should consider pulling up the sawtooth connectivity that can be shared between the event processor and the rest API.

eloaverona: I'll work on the shutdown logic for this.  And yes, it would be good to be able to share the connection between the the event processor and the rest API. Earlier, I misspoke when I said it would be easy with this current implementation. It is easy to create new message types and implement more handlers for the zmq actor, like [here](https://github.com/hyperledger/grid/blob/1c05fc11e4271fe5d982ee7aaaa35cf6a8354d0c/daemon/src/rest_api/route_handler.rs#L63). But that would not help with sharing connections with the event processor.

daemon/src/rest_api/error.rs
```diff
@@ -35,6 +47,16 @@ impl fmt::Display for RestApiError {
         match self {
             RestApiError::StartUpError(e) => write!(f, "Start-up Error: {}", e),
             RestApiError::StdError(e) => write!(f, "Std Error: {}", e),
+            RestApiError::BadRequest(ref s) => write!(f, "Bad Request: {}", s),
```
peterschwarz: I think this should be split into a `RestApiServerError` (for errors related to starting/shutting down), and `RestApiResponseError`

daemon/src/rest_api/route_handler.rs
```diff
@@ -34,6 +46,86 @@ impl SawtoothMessageSender {
...
+        .map_err(|err| {
+            RestApiError::RequestHandlerError(format!(
+                "Failed to parse validator response from bytes. {}",
+                err.to_string()
+            ))
+        })?;
+
+        match process_validator_response(response_status.get_status()) {
+            Ok(_) => {
+                let mut batch_query = String::new();
```
peterschwarz: This can be reduced to:
```
    let batch_query = msg.batch_list
        .get_batches()
        .iter()
        .map(|batch| batch.get_header_signature())
        .collect::<Vec<_>>()
        .join(",");
```

daemon/src/rest_api/route_handler.rs
```diff
@@ -34,6 +46,86 @@ impl SawtoothMessageSender {
...
+                for (i, batch) in msg.batch_list.get_batches().iter().enumerate() {
+                    batch_query.push_str(batch.get_header_signature());
+                    if msg.batch_list.get_batches().len() > 1
+                        && i <= msg.batch_list.get_batches().len() - 2
+                    {
+                        batch_query.push(',');
+                    }
+                }
+                let encoded_query: String = byte_serialize(batch_query.as_bytes()).collect();
```
peterschwarz: Not sure I understand the need for this. Its already a comma-delimited hex string, which is url-safe

daemon/src/sawtooth_connection.rs
```diff
@@ -20,17 +20,14 @@ use sawtooth_sdk::messaging::zmq_stream::{ZmqMessageConnection, ZmqMessageSender
 
 pub struct SawtoothConnectionHandler {
     sender: ZmqMessageSender,
-    _receiver: MessageReceiver,
+    pub receiver: MessageReceiver,
```
peterschwarz: Perhaps guard this behind a an accessor: `fn get_receiver(&sefl) -> &MessageReceiver`

daemon/src/rest_api/route_handler.rs
peterschwarz: Even better, 
```
...
        .map(Batch::get_header_signature)
...
```
as long as `Batch` is in your `use` statements

daemon/src/rest_api/mod.rs
```diff
@@ -52,9 +67,13 @@ pub fn run(
...
         .spawn(move || {
             let sys = actix::System::new("Grid-Rest-API");
+            let zmq_connection_addr =
+                SawtoothMessageSender::create(move |_ctx: &mut Context<SawtoothMessageSender>| {
+                    SawtoothMessageSender::new(zmq_sender)
+                });
 
             info!("Starting Rest API at {}", &bind_url);
-            let addr = server::new(create_app)
+            let addr = server::new(move || create_app(zmq_connection_addr.clone()))
```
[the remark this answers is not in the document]
eloaverona: I get a compiler error without the clone there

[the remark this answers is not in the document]
peterschwarz: Ah, it might change what type the closure is (Fn, vs FnOnce vs FnMut)

daemon/src/sawtooth_connection.rs
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+
+use sawtooth_sdk::messaging::stream::{MessageConnection, MessageReceiver, MessageSender};
+use sawtooth_sdk::messaging::zmq_stream::{ZmqMessageConnection, ZmqMessageSender};
+
+pub struct SawtoothConnectionHandler {
```
peterschwarz: Handler is a bit of an overloaded term.  I would just call this thing a "SawtoothConnection"

PR #31, opened 2019-04-02
Convert TnT dev docker file to multi-stage

peterschwarz:
Convert the Track and Trace TP's development Dockerfile to use the same multi-stage build style as the grid daemon.  This improves overall build time and developer experience when working in a docker-compose environment.

### DIFF
--- a/contracts/track_and_trace/Dockerfile
+++ b/contracts/track_and_trace/Dockerfile
@@ -12,56 +12,63 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM rust:1
+FROM ubuntu:bionic as GRID-TNT-BUILDER
 
-RUN apt-get update && apt-get install -y unzip libzmq3-dev
+# Install base dependencies
+RUN apt-get update \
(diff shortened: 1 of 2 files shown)


contracts/track_and_trace/Dockerfile
```diff
@@ -12,56 +12,67 @@
...
-RUN apt-get update && apt-get install -y protobuf-compiler
+    && unzip -o protoc-3.5.1-linux-x86_64.zip -d /usr/local \
+    && rm protoc-3.5.1-linux-x86_64.zip
+
+ENV PATH=$PATH:/protoc3/bin:/root/.cargo/bin
 
 # Build TP with dummy source in order to cache dependencies in Docker image.
 # Make sure not to use the `volumes` command to overwrite:
 #   - /grid/contracts/track_and_trace/target/
 #   - /grid/contracts/track_and_trace/src/messages/
```
agunde406: the line about messages can also be removed

peterschwarz: Removed

PR #32, opened 2019-04-03
Configure Grid daemon to write events to a PostGreSQL instance

peterschwarz:
This currently writes blocks to a Block table, but does not (yet) deal in fork resolution.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -93,7 +93,17 @@ services:
       - 8080
     ports:
       - '8080:8080'
-    entrypoint: /gridd -vv -b gridd:8080 -C tcp://validator:4004
+    entrypoint: |
+        bash -c "
+          # we need to wait for the db to have started.
+          until PGPASSWORD=grid_example psql -h db -U grid -c '\q'; do
+              >&2 echo \"Database is unavailable - sleeping\"
+              sleep 1
+          done
+
+          /gridd -vv -b gridd:8080 -C tcp://validator:4004 \
+              --database-url postgres://grid:grid_example@db/grid
+        "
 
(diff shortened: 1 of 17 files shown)


peterschwarz:
To test this after running `docker-compose up --build`, you can run

```
$ docker run -it --rm --network grid_default  postgres psql -h db -U grid
Password for user grid: < type "grid_example" >
psql (11.2 (Debian 11.2-1.pgdg90+1))
Type "help" for help.

grid=# select * from block;
                                                             block_id                                                             | block_num |                         state_root_hash
----------------------------------------------------------------------------------------------------------------------------------+-----------+------------------------------------------------------------------
 8bbf805bc297ee264c9aaae43f79c6730c4909be9766db0c9e74ef7b53eee24e58a46e41c14174385c0d1383a1ce35b271a0a230c965904a29fe35bb89db7679 |         0 | f5f0adfe399e64c6f22ccf678409e8a9b2e05ef704eacb74217ee19dcce08bfa
(1 row)
grid=#
```

peterschwarz:
Added block num index as part of initial creation

docker-compose.yaml
```diff
@@ -93,7 +93,13 @@ services:
       - 8080
     ports:
       - '8080:8080'
-    entrypoint: /gridd -vv -b gridd:8080 -C tcp://validator:4004
+    entrypoint: |
+        bash -c "
+          # we need to wait for the db to have started.
+          sleep 10 &&
```
jsmitchell: As an alternative: https://docs.docker.com/compose/startup-order/

peterschwarz: Added

PR #33, opened 2019-04-08
Add Grid Schema Smart contract  

agunde406:
This PR implements the Smart contract described in
the Grid Primitives RFC
hyperledger/grid-rfcs#4

Note this does not include the permission check of if
the agent has the permission can_create_schema for the
organization. This we will be added in a separate PR
after the implementation of Permission checker in the
Grid SDK.

### DIFF
--- a/contracts/schema/src/payload.rs
+++ b/contracts/schema/src/payload.rs
@@ -1,0 +1,210 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use sabre_sdk::ApplyError;
+    } else {
+        use sawtooth_sdk::processor::handler::ApplyError;
+    }
+}
+
+use grid_sdk::protocol::schema::payload::{
+    Action, SchemaCreateAction, SchemaPayload, SchemaUpdateAction,
+};
+
+pub fn validate_payload(payload: &SchemaPayload) -> Result<(), ApplyError> {
+    match payload.action() {
+        Action::SchemaCreate => validate_schema_create_action(payload.schema_create()),
+        Action::SchemaUpdate => validate_schema_update_action(payload.schema_update()),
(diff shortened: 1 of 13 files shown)


contracts/schema/src/payload.rs
```diff
@@ -0,0 +1,212 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use sabre_sdk::ApplyError;
+    } else if #[cfg(test)]{
+        use sawtooth_sdk::processor::handler::ApplyError;
```
peterschwarz: Is this different from the else branch?  I'm not seeing it, if so

agunde406: Whoops, that whole test section should be removed. This is left over from before the updated sabre.

agunde406: Fixed

contracts/schema/src/state.rs
```diff
@@ -0,0 +1,419 @@
...
+                        return Ok(Some(schema.clone()));
+                    }
+                }
+                Ok(None)
+            }
+            None => Ok(None),
+        }
+    }
+
+    /// Sets a Grid Schema in state. Handles creating a SchemaList if one does not already exists
```
jsmitchell: s/exists/exist/

contracts/schema/src/state.rs
```diff
@@ -0,0 +1,419 @@
...
+    fn test_get_organization_none() {
+        let mut transaction_context = MockTransactionContext::default();
+        let mut state = GridSchemaState::new(&mut transaction_context);
+
+        let result = state.get_organization("test_org").unwrap();
+        assert!(result.is_none())
+    }
+
+    #[test]
+    // Test that if an org does exists in state, Some(Organization) is returned;
```
jsmitchell: trailing semicolon

contracts/schema/src/state.rs
```diff
@@ -0,0 +1,419 @@
...
+        let result = state.get_organization("test_org").unwrap();
+        assert!(result.is_some());
+        let organization = result.unwrap();
+
+        assert_eq!(organization.org_id(), "test_org");
+    }
+
+    #[test]
+    // 1. Test that if a schema is not in state a None is returned.
+    // 2. Test that a schema can be addec to state using set_state.
```
jsmitchell: s/addec/added/

contracts/schema/src/state.rs
agunde406: updated

contracts/schema/src/state.rs
agunde406: updated

contracts/schema/src/state.rs
agunde406: updated

PR #34, opened 2019-04-08
 Update Grid Track and Trace to be Sabre compatible

agunde406:
The changes include updating the smart contract to use
Sawtooth Rust SDK 0.3, changes needed to compile into
WASM, using sabre sdk, and adding the sabre entrypoint.

Also includes splitting out the payload and state objects into there own file.


contracts/track_and_trace/src/payload.rs
```diff
@@ -0,0 +1,139 @@
...
+                Action::CreateRecord(create_record.clone())
+            }
+            SCPayload_Action::FINALIZE_RECORD => {
+                Action::FinalizeRecord(payload.get_finalize_record().clone())
+            }
+            SCPayload_Action::CREATE_RECORD_TYPE => {
+                let create_record_type = payload.get_create_record_type();
+                if create_record_type.get_name() == "" {
+                    return Err(ApplyError::InvalidTransaction(String::from(
+                        "Record Type name cannot be an empty string",
```
jsmitchell: This error message capitalizes "Record Type", but the message on line 91 uses "Record type"

agunde406: updated to use "Record type"

contracts/track_and_trace/src/handler.rs
```diff
@@ -1245,6 +1244,7 @@ impl SupplyChainTransactionHandler {
                     Proposal_Role::OWNER => {
                         if owner.get_agent_id() != current_proposal.get_issuing_agent() {
                             current_proposal.status = Proposal_Status::CANCELED;
+                            #[cfg(not(target_arch = "wasm32"))]
                             info!("Record owner does not match the issuing agent of the proposal");
```
peterschwarz: We could supply some no-op macros if not wasm, so as not to repeat this `cfg` statement before every log message.

agunde406: I think the more correct solution would be to provide these macros in the Sabre sdk itself. I will make a story for this and update in a later PR.

PR #35, opened 2019-04-08
Connect rest API to database & get_agents route

shannynalayna:
Adds a connection for the rest API to the database created in pr #32. 

This pr is now dependent on #43 which adds the necessary agents table for this api route.

For testing, run `docker-compose.yaml` in the `daemon/test` directory. Then, run `cargo test --features test-api`

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -178,6 +178,36 @@ paths:
           description: API is unable to reach the database
           schema:
             $ref: "#/definitions/Error"
+  /agent:
+    get:
+      tags:
+        - "agents"
+      summary: "Get a list of Agents"
+      description: "Fetches a list of agents from the reporting database"
+      operationId: "list_agents"
+      produces:
+        - "application/json"
+      responses:
+        200:
+          description: "List of Agents"
+          schema:
+            properties:
+              data:
+                type: array
+                items:
+                  $ref: "#/definitions/Agents"
+        400:
+          $ref: "#/responses/400BadRequest"
+        429:
+          $ref: "#/responses/429TooManyRequests"
+        500:
+          description: Something went wrong within the database
+          schema:
+            $ref: "#/definitions/Error"
+        503:
+          description: API is unable to reach the database
+          schema:
+            $ref: "#/definitions/Error"
 
 responses:
   400BadRequest:
@@ -390,3 +420,30 @@ definitions:
         - ENUM
         - STRUCT
         - LOCATION
+
+  Agent:
+    properties:
+      public_key:
+        type: string
+        example: 038bba5708acc262464c9fe30d3de9e905a9a5fa30cedd151dd9cd09ea26d46d00
+      org_id:
+        type: string
+        example: 03c360a46457a284793153e09840c322bee1dcc34445d7d49e19e60abd18fd0758
+      active:
+        type: boolean
+      roles:
+        type: array
+        items:
+          type: string
+      metadata:
+        type: array
+        items:
+          $ref: '#definitions/Metadata'
+
+  Metadata:
+    type: object
+    properties:
+      key:
+        type: string
+      value:
+        type: string
--- a/daemon/src/database/schema.rs
+++ b/daemon/src/database/schema.rs
@@ -56,9 +56,4 @@ table! {
     }
 }
 
-allow_tables_to_appear_in_same_query!(
-    agent,
-    block,
-    chain_record,
-    organization,
-);
+allow_tables_to_appear_in_same_query!(agent, block, chain_record, organization,);
--- a/daemon/src/database/schema.rs
+++ b/daemon/src/database/schema.rs
@@ -18,13 +18,13 @@
 table! {
     agent (id) {
         id -> Int8,
-        start_block_num -> Int8,
-        end_block_num -> Int8,
         public_key -> Varchar,
         org_id -> Varchar,
         active -> Bool,
         roles -> Array<Varchar>,
         metadata -> Array<Json>,
+        start_block_num -> Int8,
+        end_block_num -> Int8,
     }
 }
 
--- a/daemon/src/rest_api/error.rs
+++ b/daemon/src/rest_api/error.rs
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::database::DatabaseError;
+
 use actix::MailboxError;
 use actix_web::error::{PayloadError, ResponseError, UrlGenerationError};
 use actix_web::HttpResponse;
+use diesel;
 use std::error::Error;
 
 use std::fmt;
@@ -55,6 +58,7 @@ pub enum RestApiResponseError {
     SawtoothConnectionError(String),
     SawtoothValidatorResponseError(String),
     RequestHandlerError(String),
+    DatabaseError(String),
 }
 
 impl Error for RestApiResponseError {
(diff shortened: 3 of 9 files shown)


daemon/src/rest_api/route_handler.rs
```diff
@@ -221,3 +221,16 @@ impl Handler<GetAgents> for DbExecutor {
         }
     }
 }
+
+pub fn get_agents(
```
dplumb94: I like to use a "list" prefix for retrieving a whole collection and "fetch" for retrieving a single resource as it eliminates some confusion and overloading of "get" (e.g. `list_agents` in this case, and `fetch_agent` for a call like `GET /agents/asdf/`). Just my personal preference though.

daemon/src/database/models.rs
```diff
@@ -24,3 +26,13 @@ pub struct Block {
     pub block_num: i64,
     pub state_root_hash: String,
 }
+
+#[derive(Insertable, Queryable, Serialize, Deserialize, Debug)]
+#[table_name = "agents"]
+pub struct Agent {
+    pub public_key: String,
```
dplumb94: These fields have a different order compared to the protobufs. I don't think it will matter here since it is consistent throughout the Rust code? I have gotten strange diesel erorrs in the past because field ordering did not match up.

daemon/src/rest_api/route_handler.rs
```diff
@@ -367,6 +367,29 @@ fn process_batch_status_response(
...
+
+impl Message for ListAgents {
+    type Result = Result<Vec<Agent>, RestApiResponseError>;
+}
+
+impl Handler<ListAgents> for DbExecutor {
+    type Result = Result<Vec<Agent>, RestApiResponseError>;
+
+    fn handle(&mut self, _msg: ListAgents, _: &mut SyncContext<Self>) -> Self::Result {
+        if let Ok(agents) = db::get_agents(&*self.connection_pool.get().map_err(|err| {
```
RyanLassigBanks: You could probably just call `db::get_agents(..).map_err(..)`, since get_agents will return an empty list if there are not agents in the database.

daemon/src/database/models.rs
```diff
@@ -41,7 +42,7 @@ pub struct NewAgent {
     pub end_block_num: i64,
 }
 
-#[derive(Queryable, Debug)]
+#[derive(Queryable, Deserialize, Serialize, Debug)]
```
peterschwarz: We don't necessarily want to expose things like the internal record id nor the start and end block numbers from the api.  We should create a light-weight serializable result object for these.

daemon/openapi.yaml
```diff
@@ -178,6 +178,36 @@ paths:
...
+        - "application/json"
+      responses:
+        200:
+          description: "List of Agents"
+          schema:
+            properties:
+              data:
+                type: array
+                items:
+                  $ref: "#/definitions/Agents"
```
peterschwarz: There's no definition of Agents in this file.

daemon/src/database/schema.rs
```diff
@@ -18,13 +18,13 @@
...
     agent (id) {
         id -> Int8,
-        start_block_num -> Int8,
-        end_block_num -> Int8,
         public_key -> Varchar,
         org_id -> Varchar,
         active -> Bool,
         roles -> Array<Varchar>,
         metadata -> Array<Json>,
+        start_block_num -> Int8,
```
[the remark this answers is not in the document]
shannynalayna: The re-ordering was originally prompted by the fact that this didn't match the Agent model definition, so diesel wasn't able to load the Agent because of this.

daemon/src/rest_api/error.rs
```diff
@@ -81,6 +85,9 @@ impl fmt::Display for RestApiResponseError {
             RestApiResponseError::RequestHandlerError(ref s) => {
                 write!(f, "Sawtooth Validator Response Error: {}", s)
             }
+            RestApiResponseError::DatabaseError(ref s) => {
+                write!(f, "Database Connection Error: {}", s)
+            }
```
eloaverona: You should also implement a mapping from `DatabaseError` to `HttpResponse` in the method below (line 96-108). Currently the `DatabaseError` is mapped to a `HttpResponse::InternalServerError`, which is not what we want I think. 
It might be a good idea to implement a `RestApiResponseError::NotFound` error that is separate from the `DatabaseError`. This error would be returned when an user tries to fetch an agent by id, but the id is not found. 
A `RestApiResponseError::NotFound` could be mapped to a `HttpResponse::NotFound`. 
Other `DatabaseErrors` might be better mapped to `HttpResponse::InternalServerError` or               `HttpResponse::ServiceUnavailable`.

daemon/src/database/models.rs
```diff
@@ -56,6 +57,15 @@ pub struct Agent {
     pub end_block_num: i64,
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+pub struct ReturnAgent {
```
peterschwarz: This is specific to the rest api's output, and shouldn't live here in the database models.

Likewise, I think you could have it just be references.  See this playground for an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e93e6dca7a540974d1feb74bcb6abc54

daemon/src/rest_api/route_handler.rs
```diff
@@ -352,11 +367,45 @@ fn process_batch_status_response(
...
+
+impl Message for ListAgents {
+    type Result = Result<Vec<ReturnAgent>, RestApiResponseError>;
+}
+
+impl Handler<ListAgents> for DbExecutor {
+    type Result = Result<Vec<ReturnAgent>, RestApiResponseError>;
+
+    fn handle(&mut self, _msg: ListAgents, _: &mut SyncContext<Self>) -> Self::Result {
+        Ok(db::get_agents(&*self.connection_pool.get()?)?)
```
eloaverona: You probably should check here if the error returned is type diesel::result::Error::NotFound and have it mapped to RestApiResponseError::ResultError while other diesel::result::Error are probably better if mapped to RestApiResponseError::DatabaseError, as they most likely are errors caused by programmer or database errors than the user's action.
http://docs.diesel.rs/diesel/result/enum.Error.html

daemon/src/database/models.rs
shannynalayna: The first part of your comment makes sense, however, I'm running into issues with using the lifetime parameter while making the struct all references since that abridged agent type would be used in the associated type for both the Message and Handler traits from actix. A lot of discussions about this issue solved it mainly by adding the lifetime to the trait directly. Not quite sure how to work around that issue, thoughts?

daemon/src/rest_api/route_handler.rs
shannynalayna: Right now, the `get_agents` function won't return a NotFound error since it uses `.load(conn)`.. Also, in the db helpers we can easily map that NotFound error to just Ok(None) like it's done here: https://github.com/hyperledger/grid/blob/master/daemon/src/database/helpers/blocks.rs#L62

eloaverona: I agree, the NotFound error should returned in the fetch_agent by public key when no agent is found with the requested public key. 
Though I still think the other diesel::result::Error enum options should not be mapped to RestApiResponseError::ResultError but rather RestApiResponseError::DatabaseError then.

shannynalayna: Yeah, I guess the hope was that the ResultError could be mapped to the NotFound HttpResponse error. May have to look back at implementing that error when it actually comes up.

PR #36, opened 2019-04-08
Batch statuses endpoint

eloaverona:
Implement GET batch_statuses endpoint as defined in [openapi spec documentation](https://github.com/hyperledger/grid/blob/440d3b4e0b59d215063cdabd2342f7f32fe47c21/daemon/openapi.yaml#L66).

### DIFF
--- a/daemon/src/rest_api/route_handler.rs
+++ b/daemon/src/rest_api/route_handler.rs
@@ -18,14 +18,15 @@ use actix::{Actor, Context, Handler, Message};
 use actix_web::{HttpMessage, HttpRequest, HttpResponse, State};
 use futures::future;
 use futures::future::Future;
-use protobuf;
 use sawtooth_sdk::messages::batch::{Batch, BatchList};
 use sawtooth_sdk::messages::client_batch_submit::{
     ClientBatchSubmitRequest, ClientBatchSubmitResponse, ClientBatchSubmitResponse_Status,
+    ClientBatchStatus, ClientBatchStatusRequest, ClientBatchStatusResponse,
 };
 use sawtooth_sdk::messages::validator::Message_MessageType;
 use sawtooth_sdk::messaging::stream::MessageSender;
 use serde::Serialize;
+use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


daemon/src/rest_api/route_handler.rs
```diff
@@ -158,10 +208,95 @@ pub fn submit_batches(
...
+        }
+    };
+    let wait = match query.get("wait") {
+        Some(wait_time) => {
+            if wait_time == "false" {
+                None
+            } else {
+                match wait_time.parse::<u32>() {
+                    Ok(wait_time) => Some(wait_time),
+                    Err(_) => Some(DEFAULT_TIME_OUT),
```
vaporos: So, i fI set "wait=a", this will set it to the default. Shouldn't we error out because it's an invalid request?

daemon/src/rest_api/route_handler.rs
```diff
@@ -55,6 +57,53 @@ impl Message for SubmitBatches {
...
+                .iter()
+                .map(|txn| {
+                    let mut invalid_transaction_info = HashMap::new();
+                    invalid_transaction_info
+                        .insert("id".to_string(), txn.get_transaction_id().to_string());
+                    invalid_transaction_info
+                        .insert("message".to_string(), txn.get_message().to_string());
+                    invalid_transaction_info.insert(
+                        "extended_data".to_string(),
+                        hex::encode(txn.get_extended_data()),
```
peterschwarz: This should be Base64-encoded, to match the Sawtooth API

daemon/src/rest_api/route_handler.rs
```diff
@@ -212,6 +212,43 @@ pub fn get_batch_statuses(
     unimplemented!()
 }
 
+fn query_validator<T: protobuf::Message, C: protobuf::Message>(
```
agunde406: Can you add a comment to the commit about the addition of this method.

eloaverona: Addressed

PR #38, opened 2019-04-10
Remove pike example directory

RyanLassigBanks:
Removes pike examples directory and updates Sabre SDK dependency for the Pike TP. The example directory will be moved to grid-contrib. This update was necessary because broken paths in examples directory make it impossible to import grid-sdk


RyanLassigBanks:
> Let's not fix stuff directly in the sawtooth-pike directory. Instead, fix it in the PR that moves it to its final location.

Normally I would agree, but the examples would depend on the Grid SDK. I can't pull or compile the Grid SDK, because the paths in several of the pike modules point to locations outside of the repo. So in order to put up a PR with the example set up correctly and compilable I need to fix the paths in Grid (hence this PR). I suppose an option would be to copy over the grid protobufs along with the examples, but that seems bad.

PR #41, opened 2019-04-11
Add Pike TP to Grid SDK contracts

dplumb94:
Moves the Pike TP from the sawtooth-pike directory to grid/contracts. Also updates the Pike contract code to use the Grid SDK protos, as well as the updated APIs for Sabre and the Rust SDK. The code from the `addresser` crate was added as a module to the contract code.

The Dockerfiles for Pike were brought in line with the track and trace Dockerfiles, and Grid docker-compose files were updated to include Pike.

PR #42, opened 2019-04-12
Add metadata to organizations

dplumb94:
Based on #41

### DIFF
--- a/docs/source/pike_transaction_family.rst
+++ b/docs/source/pike_transaction_family.rst
@@ -68,9 +68,10 @@ Organization
 
 An organization has three fields:
 
-- id: a unique identifier for the organization
-- name: a user defined identifier for the organization
-- address: a physical address for the organization
+- id: A unique identifier for the organization.
+- name: A user defined identifier for the organization.
+- address: A physical address for the organization.
+- metadata: A list of key value pairs describing data about the organization.
 
 The id is the unique key for an Organization.
 
@@ -80,6 +81,7 @@ The id is the unique key for an Organization.
         string org_id = 1;
         string name = 2;
         string address = 3;
+        repeated KeyValueEntry metadata = 4;
     }
 
 Organization List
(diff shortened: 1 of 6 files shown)


sdk/protos/pike_payload.proto
```diff
@@ -56,10 +56,12 @@ message CreateOrganizationAction {
   string id = 1;
   string name = 2;
   string address = 3;
+  repeated KeyValueEntry metadata = 4;
```
agunde406: This also needs to be added to protocol::pike::payload.

sdk/protos/pike_payload.proto
```diff
@@ -56,10 +56,12 @@ message CreateOrganizationAction {
...
 }
 
 message UpdateOrganizationAction {
   string id = 1;
   string name = 2;
   string address = 3;
+  repeated KeyValueEntry metadata = 4;
```
agunde406: This also needs to be added to protocol::pike::payload.

sdk/protos/pike_state.proto
```diff
@@ -36,6 +36,7 @@ message Organization {
   string org_id = 1;
   string name = 2;
   string address = 3;
+  repeated KeyValueEntry metadata = 4;
```
agunde406: This also needs to be added to protocol::pike::state.

sdk/protos/pike_payload.proto
dplumb94: Updated payload and state to add metadata to orgs in a new commit

docs/source/pike_transaction_family.rst
```diff
@@ -71,6 +71,7 @@ An organization has three fields:
 - id: a unique identifier for the organization
 - name: a user defined identifier for the organization
 - address: a physical address for the organization
+- metadata: A list of key value pairs describing data about the organization.
```
chenette: Extremely minor nitpick: Should be consistent with the case and non-punctuation of the other entries (start with a lower-case a and remove the period).

dplumb94: I will fix that! Good catch

PR #44, opened 2019-04-18
Implement Pike PermissionChecker in Grid SDK

dplumb94:
PermissionChecker is a helper that can be used to check if a Pike agent has a certain permission. It can be imported from the Grid SDK into a contract to execute Pike functionality. This implementation is fairly simple, but it can be extended to allow smart permission checks later on.

In order to test the PermissionChecker, I moved the MockTransactionContext implementation from the Schema contract into the Grid SDK.  This can be used to test other contracts or SDK functionality that requires a TransactionContext. MockTransactionContext is backed by a HashMap contained in a RefCell. Due to the properties of RefCell, It should only be used in unit tests.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -22,6 +22,15 @@ edition = "2018"
 
 [dependencies]
 protobuf = "2"
+cfg-if = "0.1"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+rust_crypto = {git = "https://github.com/agunde406/rust-crypto", branch="wasm_sha2"}
+sabre-sdk = "0.2"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+rust-crypto = "0.2"
+sawtooth-sdk = {git = "https://github.com/hyperledger/sawtooth-sdk-rust"}
 
 [build-dependencies]
 protoc-rust = "2"
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -12,5 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[macro_use]
+extern crate cfg_if;
+
+pub mod permissions;
 pub mod protocol;
 pub mod protos;
--- a/sdk/src/permissions.rs
+++ b/sdk/src/permissions.rs
@@ -1,0 +1,323 @@
+// Copyright 2019 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crypto::digest::Digest;
+use crypto::sha2::Sha512;
+use std::error::Error;
+use std::fmt;
+
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use sabre_sdk::WasmSdkError as ContextError;
+        use sabre_sdk::TransactionContext;
+    } else {
+        use sawtooth_sdk::processor::handler::ContextError;
+        use sawtooth_sdk::processor::handler::TransactionContext;
+    }
+}
+
+use crate::protos::{FromBytes, ProtoConversionError};
+use crate::protocol::pike::state::{Agent, AgentList};
+
+const PIKE_NAMESPACE: &'static str = "cad11d";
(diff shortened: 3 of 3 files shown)


sdk/src/mock.rs
```diff
@@ -0,0 +1,71 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use sabre_sdk::{TransactionContext, ContextError};
```
agunde406: If we make this only available during tests, we do not need to pull in saber here as the tests will not be run when compiling into wasm.

sdk/src/lib.rs
```diff
@@ -12,5 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[macro_use]
+extern crate cfg_if;
+
+pub mod mock;
```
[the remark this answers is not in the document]
agunde406: This doesn't actually work the way I was thinking it did since schema needs to pull this in from grid sdk. It might make sense to duplicate the MockTransactionContext in both places afterall.

sdk/src/permissions.rs
```diff
@@ -0,0 +1,261 @@
...
+pub struct PermissionChecker<'a> {
+    context: &'a TransactionContext,
+}
+
+impl<'a> PermissionChecker<'a> {
+    pub fn new(context: &'a TransactionContext) -> PermissionChecker {
+        PermissionChecker { context: context }
+    }
+
+    pub fn has_permission(
```
agunde406: Can you please add a doc comment to has_permission

sdk/src/permissions.rs
```diff
@@ -0,0 +1,261 @@
...
+        self, public_key: &str, role: &str
```
agunde406: Let's call `role` here `permission`.  We are currently overloading the roles in agent to mean permissions.

contracts/schema/src/handler.rs
```diff
@@ -232,71 +232,27 @@ fn schema_update(
...
     use sawtooth_sdk::processor::handler::ApplyError;
-    use sawtooth_sdk::processor::handler::{ContextError, TransactionContext};
 
     use crate::state::{compute_agent_address, compute_schema_address};
 
-    #[derive(Default)]
-    /// A MockTransactionContext that can be used to test GridSchemaState
-    struct MockTransactionContext {
-        state: RefCell<HashMap<String, Vec<u8>>>,
+    //  Implements functions to add to things state for testing
```
agunde406: to add to things state -> to add things to state

agunde406: This typo has not been fixed

sdk/src/lib.rs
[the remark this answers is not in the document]
vaporos: I agree, mock stuff should not be part of the grid sdk.

dplumb94:
> As this is library code, we need API doc on all the public methods and examples of how to use it.

Updated to include better documentation on the public functions, errors, and structs.

What do you mean by examples?

sdk/Cargo.toml
```diff
@@ -22,6 +22,15 @@ edition = "2018"
 
 [dependencies]
 protobuf = "2"
+cfg-if = "0.1"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+rust_crypto = {git = "https://github.com/agunde406/rust-crypto", branch="wasm_sha2"}
+sabre-sdk = "0.2"
+
+[target.'cfg(unix)'.dependencies]
```
vaporos: Can this be !wasm32? Because otherwise, this definitely won't work on Windows right?

sdk/Cargo.toml
```diff
@@ -22,6 +22,15 @@ edition = "2018"
...
+rust-crypto = "0.2.36"
```
vaporos: This should be "0.2".

sdk/src/permissions.rs
```diff
@@ -0,0 +1,331 @@
...
+// limitations under the License.
+
+use crypto::digest::Digest;
+use crypto::sha2::Sha512;
+use std::error::Error;
+use std::fmt;
+
+cfg_if! {
+    if #[cfg(target_arch = "wasm32")] {
+        use sabre_sdk::ContextError;
```
agunde406: this doesnt actually exist in sabre so this doesnt work. please add make sure this compiles with `cargo +nightly build --target wasm32-unknown-unknown` you will need to add `rustup target add wasm32-unknown-unknown --toolchain nightly` to be able to test this.

sdk/Cargo.toml
agunde406: Note this can be changed to `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]`  almost everywhere we do this in sabre (and in other grid code) we use unix because I thought this didnt work (I dont remember why I thought this, it was over a year ago). I will create a story to make sure it gets updated everywhere so it could work on windows.

PR #48, opened 2019-04-23
Remove some old pike files

vaporos:


vaporos:
> Pike addressor has been moved into the smart contract, not the grid SDK

Updated the commit message.

PR #49, opened 2019-04-23
Update Grid Schema Smart Contract to use the Permission Checker

dplumb94:
Based on #44 

Adds permission checks to the Schema contract. Also makes some tweaks to the PermissionChecker and GridSchemaState to cleanly integrate this functionality.

### DIFF
--- a/contracts/schema/src/handler.rs
+++ b/contracts/schema/src/handler.rs
@@ -27,6 +27,7 @@ cfg_if! {
     }
 }
 
+use grid_sdk::permissions::PermissionChecker;
 use grid_sdk::protocol::schema::payload::{
     Action, SchemaCreateAction, SchemaPayload, SchemaUpdateAction,
 };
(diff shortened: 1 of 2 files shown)


contracts/schema/src/handler.rs
```diff
@@ -147,6 +150,8 @@ fn schema_create(
         )));
     }
 
+    check_permission(perm_checker, signer, "can_create_schema")?;
```
agunde406: Can you move this check above the fetching of the agent? that means we can skip re-fetching the agent if he does not have permissions.

contracts/schema/src/handler.rs
```diff
@@ -201,6 +207,8 @@ fn schema_update(
         )));
     }
 
+    check_permission(perm_checker, signer, "can_update_schema")?;
```
agunde406: Same as above.

PR #51, opened 2019-04-23
Remove Dockerfile-installed-* for smart contracts

vaporos:
In the short term, we will be deploying these as sabre smart contracts.
So testing TP deb builds is no longer useful to us. Maintaining these
Dockerfiles has a lot of overhead, so proactively removing them.

Signed-off-by: Shawn T. Amundson <amundson@bitwise.io>


Jenkinsfile
```diff
@@ -83,7 +83,6 @@ node ('master') {
 
         stage("Archive Build artifacts") {
             sh 'mkdir -p build/debs'
-            sh 'docker run -v $(pwd)/build/debs:/build grid-track-and-trace-tp-installed:$ISOLATION_ID bash -c "cp /tmp/grid-track-and-trace-tp*.deb /build"'
             archiveArtifacts artifacts: '*.tgz, *.zip'
             archiveArtifacts artifacts: 'build/debs/*.deb'
```
rbuysse: Since we're not currently generating any artifacts, this will need to be updated so it doesn't fail when nothing matches. 

archiveArtifacts artifacts: build/debs/*.deb', allowEmptyArchive: true

PR #59, opened 2019-04-30
Add /agent/{public_key} route

shannynalayna:
Finishes the agents route with the /agent/{public_key} route which allows the rest API to fetch an agent based on the public key. 

Test with: `cargo test --features test-api --test-threads=1`


daemon/src/rest_api/routes/mod.rs
```diff
@@ -653,6 +656,48 @@ mod test {
         assert_eq!(org.address, UPDATED_ADDRESS_2.to_string());
     }
```
dplumb94: The other tests have an extra row of doc comment slashes here. Not sure if it's necessary for the docs? I'd recommend adding them for consistency if not.

shannynalayna: Addressed

PR #60, opened 2019-05-01
Rbuysse package gridd

rbuysse:


docker-compose-installed.yaml
```diff
@@ -16,6 +16,18 @@
 version: '2.1'
 
 services:
+
+  gridd:
+    image: gridd-installed:${ISOLATION_ID}
+    container_name: gridd-installed:${ISOLATION_ID}
```
agunde406: This should just be `gridd-installed` right?

agunde406: aka image has isolation id and container name does not.

PR #62, opened 2019-05-01
Track and Trace Protocol for SDK

RyanLassigBanks:
Adds track and trace protobuf definition that are compatible with Pike and Schema, as well as their corresponding native objects

### DIFF
--- a/contracts/track_and_trace/src/handler.rs
+++ b/contracts/track_and_trace/src/handler.rs
@@ -32,20 +32,32 @@ cfg_if! {
     }
 }
 
-use grid_sdk::protos::track_and_trace_agent::TrackAndTraceAgent as Agent;
-use grid_sdk::protos::track_and_trace_payload::CreateTrackAndTraceAgentAction as CreateAgentAction;
-use grid_sdk::protos::track_and_trace_payload::{
-    AnswerProposalAction, AnswerProposalAction_Response, CreateProposalAction, CreateRecordAction,
-    CreateRecordTypeAction, FinalizeRecordAction, RevokeReporterAction, UpdatePropertiesAction,
+use grid_sdk::protos::deprecated_track_and_trace_payload::DeprecatedCreateTrackAndTraceAgentAction as CreateAgentAction;
+use grid_sdk::protos::deprecated_track_and_trace_payload::{
+    DeprecatedAnswerProposalAction as AnswerProposalAction,
+    DeprecatedAnswerProposalAction_Response as AnswerProposalAction_Response,
+    DeprecatedCreateProposalAction as CreateProposalAction,
+    DeprecatedCreateRecordAction as CreateRecordAction,
+    DeprecatedCreateRecordTypeAction as CreateRecordTypeAction,
+    DeprecatedFinalizeRecordAction as FinalizeRecordAction,
+    DeprecatedRevokeReporterAction as RevokeReporterAction,
+    DeprecatedUpdatePropertiesAction as UpdatePropertiesAction,
 };
+use grid_sdk::protos::track_and_trace_agent::TrackAndTraceAgent as Agent;
 use grid_sdk::protos::track_and_trace_property::{
-    Property, PropertyPage, PropertyPage_ReportedValue, PropertySchema, PropertySchema_DataType,
-    Property_Reporter, TrackAndTracePropertyValue,
+    DeprecatedProperty as Property, DeprecatedPropertyPage as PropertyPage,
+    DeprecatedPropertyPage_ReportedValue as PropertyPage_ReportedValue,
+    DeprecatedProperty_Reporter as Property_Reporter, PropertySchema, PropertySchema_DataType,
+    TrackAndTracePropertyValue,
 };
 use grid_sdk::protos::track_and_trace_proposal::{
-    Proposal, ProposalContainer, Proposal_Role, Proposal_Status,
+    DeprecatedProposal as Proposal, DeprecatedProposalContainer as ProposalContainer,
+    DeprecatedProposal_Role as Proposal_Role, DeprecatedProposal_Status as Proposal_Status,
 };
-use grid_sdk::protos::track_and_trace_record::{Record, RecordType, Record_AssociatedAgent};
+use grid_sdk::protos::track_and_trace_record::{
+    DeprecatedRecord as Record, DeprecatedRecord_AssociatedAgent as Record_AssociatedAgent,
+    RecordType,
+};
 
 use crate::addressing::*;
 use crate::payload::{Action, SupplyChainPayload};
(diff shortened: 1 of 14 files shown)


peterschwarz:
Those macros are cool. Can they be more broadly used for the other Native <-> proto conversions?

RyanLassigBanks:
> Those macros are cool. Can they be more broadly used for the other Native <-> proto conversions?

Thank you! I spent a lot of time thinking about this. I had a couple of ideas of how to do it, but I couldn't quite come up with a syntax for the macro that felt good to use and wasn't overly complicated. I would have liked to spend more time thinking about it, but I felt moving the other TnT stories forward this sprint was a higher priority.

RyanLassigBanks:
> What data type does this default result in?

The default for `PropertyDefinition` Would be this

```
PropertyDefinition {
    name: String::new(),
    data_type: DataType::Bytes,
    required: false,
    number_exponent: 0,
    enum_options: Vec::new(),
    struct_properties: Vec::new()
}
```

RyanLassigBanks:
> This has a default implementation, which just returns `None`. Not really necessary until there are sources to return.

Fair. I remove the source impl then

sdk/protos/deprecated_track_and_trace_payload.proto
```diff
@@ -0,0 +1,137 @@
+// Copyright 2017 Intel Corporation
```
agunde406: typo in commit messsage implimentatin -> implementation

sdk/src/protocol/macros.rs
```diff
@@ -0,0 +1,184 @@
+// Copyright 2019 Cargill Incorporated
```
agunde406: typo in commit message gerated -> generated

sdk/src/protocol/track_and_trace/payload.rs
```diff
@@ -0,0 +1,471 @@
...
+        match actions {
+            TrackAndTracePayload_Action::CREATE_RECORD => Ok(Action::CreateRecord),
+            TrackAndTracePayload_Action::FINALIZE_RECORD => Ok(Action::FinalizeRecord),
+            TrackAndTracePayload_Action::UPDATE_PROPERTIES => Ok(Action::UpdateProperties),
+            TrackAndTracePayload_Action::CREATE_PROPOSAL => Ok(Action::CreateProposal),
+            TrackAndTracePayload_Action::ANSWER_PROPOSAL => Ok(Action::AnswerProposal),
+            TrackAndTracePayload_Action::REVOKE_REPORTER => Ok(Action::RevokeReporter),
+            TrackAndTracePayload_Action::UNSET_ACTION => {
+                Err(ProtoConversionError::InvalidTypeError(
+                    "Cannot convert SchemaPayload_Action with type unset.".to_string(),
```
agunde406: SchemaPayload_Action -> TrackAndTracePayload_Action

sdk/src/protocol/track_and_trace/state.rs
```diff
@@ -0,0 +1,611 @@
...
+        ));
+
+        Ok(proto)
+    }
+}
+
+impl_from_bytes!(RecordList, track_and_trace_state::RecordList);
+impl_into_bytes!(RecordList);
+impl_into_proto!(RecordList, track_and_trace_state::RecordList);
+impl_into_native!(RecordList, track_and_trace_state::RecordList);
```
agunde406: Please add unit tests.

sdk/src/protocol/track_and_trace/payload.rs
```diff
@@ -0,0 +1,471 @@
...
+);
+impl_into_bytes!(TrackAndTracePayload);
+impl_into_proto!(
+    TrackAndTracePayload,
+    track_and_trace_payload::TrackAndTracePayload
+);
+impl_into_native!(
+    TrackAndTracePayload,
+    track_and_trace_payload::TrackAndTracePayload
+);
```
agunde406: Please add unit tests

sdk/src/protocol/macros.rs
```diff
@@ -0,0 +1,184 @@
...
+///         self.record_id = Some(value);
+///         self
+///     }
+///
+///     pub fn name(mut self, value: &str) -> Self {
+///         self.name = Some(value);
+///         self
+///     }
+///
+///     pub fn build(self) -> Result<FianlizeRecordAction, BuilderError> {
```
vaporos: Typo on Finalized

sdk/src/protocol/schema/state.rs
```diff
@@ -110,7 +117,7 @@ impl IntoProto<protos::schema_state::LatLong> for LatLong {}
 impl IntoNative<LatLong> for protos::schema_state::LatLong {}
 
 /// Native implementation of PropertyDefinition
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Default, Debug, Clone, PartialEq)]
```
peterschwarz: I'm not sure Default makes sense on this structure - I don't think empty name is allowed.

sdk/src/protocol/schema/state.rs
```diff
@@ -677,7 +684,7 @@ impl SchemaListBuilder {
 }
 
 /// Native implementation of PropertyValue
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Default, Debug, Clone, PartialEq)]
```
peterschwarz: I don't think `Default` makes sense here either.

sdk/src/protocol/track_and_trace/state.rs
```diff
@@ -0,0 +1,1886 @@
...
+    pub fn with_fixed(mut self, value: bool) -> Self {
+        self.fixed = Some(value);
+        self
+    }
+    pub fn with_unit(mut self, value: String) -> Self {
+        self.unit = Some(value);
+        self
+    }
+    pub fn build(self) -> Result<Property, BuilderError> {
+        let name = self.name;
```
peterschwarz: Not sure I agree with all this name-shadowing in these build methods.

contracts/track_and_trace/src/handler.rs
```diff
@@ -32,20 +32,32 @@ cfg_if! {
     }
 }
 
-use grid_sdk::protos::track_and_trace_agent::TrackAndTraceAgent as Agent;
-use grid_sdk::protos::track_and_trace_payload::CreateTrackAndTraceAgentAction as CreateAgentAction;
-use grid_sdk::protos::track_and_trace_payload::{
-    AnswerProposalAction, AnswerProposalAction_Response, CreateProposalAction, CreateRecordAction,
-    CreateRecordTypeAction, FinalizeRecordAction, RevokeReporterAction, UpdatePropertiesAction,
+use grid_sdk::protos::deprecated_track_and_trace_payload::DeprecatedCreateTrackAndTraceAgentAction as CreateAgentAction;
```
peterschwarz: Typo in commit message: depricated -> deprecated

PR #63, opened 2019-05-02
Fix Pike state delta subscription

RyanLassigBanks:
Fixes event filter for Pike state delta events

### DIFF
--- a/daemon/src/event/block.rs
+++ b/daemon/src/event/block.rs
@@ -73,6 +73,7 @@ impl EventHandler for BlockEventHandler {
                             "Fork detected. Replaced {} at height {}, with block {}.",
                             &b.block_id, &b.block_num, &block.block_id
                         );
+                        db::insert_block(&conn, &block)?;
                     }
                     Ok(Some(_)) => {
                         info!(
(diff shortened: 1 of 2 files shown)


daemon/src/event/block.rs
```diff
@@ -57,6 +57,8 @@ impl EventHandler for BlockEventHandler {
             block.block_id, block.block_num, block.state_root_hash
         );
 
+        debug!("The following operations will be performed: {:#?}", db_ops);
```
agunde406: This is removed in the next commit. If it should be removed it should be removed from this commit.

agunde406: Same with derive debug below

peterschwarz: After @agunde406  pointed this message out, I'm not sure this is a good debug message.  Just trying to think about debugging this from a sysadmin perspective - it's potentially a huge message.

RyanLassigBanks: I see your point. I would argue that debug level messages are suppose to over communicate, but I could also see it potentially being cumbersome under some circumstances. I think some logging there would be useful. What would you suggest?

RyanLassigBanks: `debug!` -> `trace!`

PR #65, opened 2019-05-03
Rest API routes for Grid Schemas

shannynalayna:
Adds the necessary functions, handlers, and implementations for the /schema and /schema/{name} routes to list and fetch grid schemas, respectively. Also adds tests for these two routes. 

To test: `cargo test --features test-api -- --test-threads=1`

### DIFF
--- a/daemon/src/database/helpers/grid_schemas.rs
+++ b/daemon/src/database/helpers/grid_schemas.rs
@@ -113,3 +113,17 @@ pub fn fetch_grid_schema(conn: &PgConnection, name: &str) -> QueryResult<Option<
         .map(Some)
         .or_else(|err| if err == NotFound { Ok(None) } else { Err(err) })
 }
+
+pub fn list_grid_property_definitions_with_schema_name(
+    conn: &PgConnection,
+    schema_name: &str,
+) -> QueryResult<Vec<GridPropertyDefinition>> {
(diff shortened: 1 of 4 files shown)


agunde406:
The above error is because this branch does not have the schema sde handling included. Just needs to be rebased.

agunde406:
Closed by accident

daemon/src/database/helpers/grid_schemas.rs
```diff
@@ -113,3 +113,17 @@ pub fn fetch_grid_schema(conn: &PgConnection, name: &str) -> QueryResult<Option<
         .map(Some)
         .or_else(|err| if err == NotFound { Ok(None) } else { Err(err) })
 }
+
+pub fn list_grid_property_definitions_with_schema_name(
```
peterschwarz: Is this used in this PR?  I can't tell if the commits are listed out of order, or if there is a commit missing

shannynalayna: Yes, it's used in this [commit](https://github.com/hyperledger/grid/pull/65/commits/a3e1fc0ab7d1de7b11144b7f709e37f3b582d869) ... it looks like the commits are listed out of order because I went back and added the commits for the property definitions after I had the initial /schema routes

PR #66, opened 2019-05-03
Fix rustfmt diff and clippy errors

dplumb94:
- [x] run rustfmt
- [x] fix clippy and rustc warnings

Adding rustfmt and clippy to CI will be in a later PR

### DIFF
--- a/contracts/track_and_trace/src/handler.rs
+++ b/contracts/track_and_trace/src/handler.rs
@@ -634,9 +634,8 @@ impl SupplyChainTransactionHandler {
         };
 
         let mut proposal_index = 0;
-        let mut count = 0;
 
-        for prop in proposals.get_entries() {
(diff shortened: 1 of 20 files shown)


contracts/track_and_trace/src/handler.rs
```diff
@@ -53,6 +53,7 @@ use crate::state::SupplyChainState;
 
 const PROPERTY_PAGE_MAX_LENGTH: usize = 256;
 
+#[derive(Default)]
```
peterschwarz: Not sure this actually makes sense here - `Default::default` will then create a handler with no family name, no versions and no namespaces, which is an invalid handler.

dplumb94: Agreed. Will remove this and add an allow for that clippy warning.

PR #69, opened 2019-05-08
Run unit and integration tests in Jenkins

eloaverona:
Updates the Jenkinsfile to run rust unit tests and integration tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -72,7 +72,17 @@ node ('master') {
         // environment doesn't need all the dependencies.
         stage("Build Test Dependencies") {
             sh 'docker-compose -f docker-compose-installed.yaml build --force-rm'
+            sh 'docker-compose -f docker/compose/grid_tests.yaml build --force-rm'
         }
+
+        stage("Run unit tests") {
+            sh 'docker-compose -f docker/compose/grid_tests.yaml up --abort-on-container-exit --exit-code-from grid_tests'
+            sh 'docker-compose -f docker/compose/grid_tests.yaml down'
+        }
+
+         stage("Run integration tests") {
+            sh './bin/run_integration_tests'
+         }
 
         stage("Create git archive") {
             sh '''
--- a/daemon/src/rest_api/routes/mod.rs
+++ b/daemon/src/rest_api/routes/mod.rs
@@ -86,7 +86,7 @@ mod test {
     use sawtooth_sdk::messaging::stream::{MessageFuture, MessageSender, SendError};
     use std::sync::mpsc::channel;
 
-    static DATABASE_URL: &str = "postgres://grid_test:grid_test@localhost:5433/grid_test";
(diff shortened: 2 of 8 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,74 @@
+# Copyright 2019 Cargill Incorporated
```
vaporos: Should we name this Dockerfile something more specific?

eloaverona: Like `tests_dockerfile`?

daemon/src/rest_api/routes/mod.rs
```diff
@@ -82,7 +82,7 @@ mod test {
     use sawtooth_sdk::messaging::stream::{MessageFuture, MessageSender, SendError};
     use std::sync::mpsc::channel;
 
-    static DATABASE_URL: &str = "postgres://grid_test:grid_test@localhost:5433/grid_test";
+    static DATABASE_URL: &str = "postgres://grid_test:grid_test@test_server:5432/grid_test";
```
agunde406: Typo in commit message. demon -> daemon

docker/Dockerfile
dplumb94: I'd suggest `tests` or `run-tests`

docker/compose/grid_tests.yaml
```diff
@@ -0,0 +1,15 @@
+version: "3.6"
```
agunde406: Missing license

docker/tests
```diff
@@ -0,0 +1,74 @@
...
+
+#Build modules with dummy source in order to cache dependencies in Docker image.
+COPY ./contracts/schema/Cargo.toml ./Cargo.toml
+RUN cargo check
+
+WORKDIR /
+RUN USER=root cargo new --bin daemon
+WORKDIR /daemon
+COPY ./daemon/Cargo.toml ./Cargo.toml
+RUN cargo check
```
rbuysse: Looks like you're running cargo check as part of building this image. Should this be a separate step in the compose file?

eloaverona: I'm running this here so that the rust dependencies can be downloaded and cached in the image.

Jenkinsfile
```diff
@@ -72,8 +72,18 @@ node ('master') {
...
+            sh 'docker-compose -f docker/compose/grid_tests.yaml build --force-rm'
         }
 
+        stage("Run unit tests") {
+            sh 'docker-compose -f docker/compose/grid_tests.yaml up --abort-on-container-exit --exit-code-from grid_tests'
+            sh 'docker-compose -f docker/compose/grid_tests.yaml down'
+        }
+
+         stage("Run integration tests") {
+             sh './bin/run_integration_tests'
```
rberg2: There is an extra space on this line

eloaverona: Addressed!

docker/tests
eloaverona: After an offline discussion, I'm marking this conversation as resolved.

PR #71, opened 2019-05-10
Add CLI commands to show and list Grid Schemas

shannynalayna:
Adds the CLI commands to list all GridSchemas and show a GridSchema based on the schema name. 

~Note: running `cargo test` in the cli directory will return an error because of a typo getting fixed up in pr #69 although no unit tests were added for these commands~
PR #69 was merged, this no longer stands

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -106,6 +106,10 @@ fn run() -> Result<(), CliError> {
             (@subcommand list =>
                 (about: "List currently defined schemas")
             )
+            (@subcommand show =>
+                (about: "Show schema specified by name argument")
+                (@arg name: +takes_value +required "Name of schema")
+            )
         )
     )
     .get_matches();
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -103,6 +103,13 @@ fn run() -> Result<(), CliError> {
                 (about: "Update schemas from a yaml file")
                 (@arg path: +takes_value +required "Path to yaml file containing a list of schema definitions")
             )
+            (@subcommand list =>
+                (about: "List schemas")
+            )
+            (@subcommand show =>
+                (about: "Show schema based on name")
```
chenette: This description needs more clarification.  What does "based on name" mean? For example, if you enter the schema name "fred", what output do you see? (Assuming, of course, that there is a schema named "fred".)

chenette: The revised version looks good!  I approve.  However, I didn't look at anything except the help-output comments, so I won't actually approve this PR officially.

cli/src/main.rs
```diff
@@ -103,6 +103,9 @@ fn run() -> Result<(), CliError> {
                 (about: "Update schemas from a yaml file")
                 (@arg path: +takes_value +required "Path to yaml file containing a list of schema definitions")
             )
+            (@subcommand list =>
+                (about: "List schemas")
```
peterschwarz: This should probably be a bit more descriptive, something like "List the currently defined schemas" or "List the schemas currently defined in global state" (or how every we are describing the chain state).

PR #75, opened 2019-05-17
Update ZMQ Dependencies

RyanLassigBanks:
Removes sawtooth-zmq dependency from schema and track_and_trace, and updates sawtooth sdk dependency in client.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -22,7 +22,7 @@ edition = "2018"
 clap = "2"
 log = "0.4"
 simple_logger = "1.0"
-sawtooth-sdk = "0.2.0"
+sawtooth-sdk = {git = "https://github.com/hyperledger/sawtooth-sdk-rust"}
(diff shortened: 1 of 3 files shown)


cli/Cargo.toml
```diff
@@ -22,7 +22,7 @@ edition = "2018"
 clap = "2"
 log = "0.4"
 simple_logger = "1.0"
-sawtooth-sdk = "0.2.0"
+sawtooth-sdk = {git = "https://github.com/hyperledger/sawtooth-sdk-rust"}
```
vaporos: Should we just have "0.3.0" here and step up the release of the sdk?

RyanLassigBanks: I would be ok with that. How fast could a release happen @rbuysse @rberg2?

RyanLassigBanks:
Waiting for 0.3 release

PR #50, opened 2019-04-23
Saberize grid

rbuysse:

PR #76, opened 2019-05-17
Updates the Grid CLI to send Sabre transactions

agunde406:
This is built on top of #50 

This includes the version of the sabre-sdk to 0.3 and updating the grid cli to send all transactions wrapped in a sabre transaction.

You can test this by running docker-compose
`docker-compose up --build`

move to the grid/cli and create a new pike organization with the following command:
`cargo run -- -vv --url http://localhost:8080 organization create org_1 OrgTest test_address`

You will see the sabre tp execute a transaction and then you can see the organization by checking hte restapi:
`http://localhost:8080/organization`

### DIFF
--- a/cli/src/transaction.rs
+++ b/cli/src/transaction.rs
@@ -23,6 +23,8 @@ use crypto::sha2::Sha512;
 use protobuf;
 use protobuf::Message;
 
+use sabre_sdk::protocol::payload::{Action, ExecuteContractActionBuilder, SabrePayloadBuilder};
+use sabre_sdk::protos::IntoBytes;
 use sawtooth_sdk::messages::batch::Batch;
 use sawtooth_sdk::messages::batch::BatchHeader;
 use sawtooth_sdk::messages::batch::BatchList;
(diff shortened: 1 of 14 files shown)


sdk/src/protocol/sabre/payload.rs
```diff
@@ -0,0 +1,3422 @@
...
+#[derive(Default, Clone)]
+pub struct CreateSmartPermissionActionBuilder {
+    name: Option<String>,
+    org_id: Option<String>,
+    function: Vec<u8>,
+}
+
+impl CreateSmartPermissionActionBuilder {
+    pub fn new() -> Self {
+        CreateSmartPermissionActionBuilder::default
```
vaporos: This and maybe some of the other structs would more naturally be an enum. (Take a look at the TnT protocol code, I think there is an example there.)

vaporos:
Should the protocol piece go into sabre, with this code pulling in the sabre SDK?

sdk/src/protocol/sabre/payload.rs
agunde406: Will take a look

agunde406:
The Sabre SDK currently is only used for writing Sabre Smart Contracts. If it was added to the sabre sdk, it should be expanded to include the state objects as well and update the rest of the sabre code to use the protocols (instead of protos) Is that what you are suggesting?

vaporos:
Yes, I think that is the logical conclusion of what I'm proposing.

cli/src/transaction.rs
```diff
@@ -150,3 +213,61 @@ fn bytes_to_hex_str(b: &[u8]) -> String {
...
+    String::from(SABRE_CONTRACT_PREFIX) + &bytes_to_hex_str(hash)[..64]
+}
+
+/// Returns a state address for a given namespace registry
+///
+/// # Arguments
+///
+/// * `namespace` - the address prefix for this namespace
+fn compute_namespace_registry_address(namespace: &str) -> Result<String, CliError> {
+    let prefix = match namespace.get(..6) {
```
peterschwarz: Not necessary to change, but you could do this as:
```
let prefix = namespace.get(..6).ok_or_else(||
    Err(CliError::UserError(format!(
        "Namespace must be at least 6 characters long: {}",
        namespace
    ))))?;
```

PR #78, opened 2019-05-20
Update TrackAndTrace Transaction Processor

eloaverona:
This PR:
- Updates the TrackAndTrace Transaction Processor to use the new Track and Trace state and payload protocols defined in the [Grid SDK](https://github.com/hyperledger/grid/tree/master/sdk/src/protocol/track_and_trace).
- Add unit tests to TrackAndTrace Transaction Processor.
- Removes methods to create_agent and create_record_type, which now are handled by the Pike and Schema TP respectively. 
- Updates the addressing methods in the TrackAndTrace Transaction Processor to create Pike agent address and Grid schema address
- Updates the TrackAndTrace State to fetch Pike agents and Schema from state, rather than the original SupplyChain Agent and RecordType.
- Renames all objects that had "SupplyChain" in the name to use "TrackAndTrace" instead.
- Removes the depracated Track and Trace proto files from the Grid SDK
- Updates  bin/run_unit_tests bash script and docker/tests dockerfile so Jenkins can
run track and trace unit tests.


contracts/track_and_trace/src/state.rs
```diff
@@ -344,45 +290,54 @@ impl<'a> SupplyChainState<'a> {
...
         let address = make_property_address(record_id, property_name, page_num);
         let d = self.context.get_state_entry(&address)?;
-        let mut property_pages = match d {
-            Some(packed) => match protobuf::parse_from_bytes(packed.as_slice()) {
-                Ok(property_pages) => property_pages,
+        let mut pages = match d {
+            Some(packed) => match PropertyPageList::from_bytes(packed.as_slice()) {
+                Ok(property_page_list) => property_page_list.property_pages().to_vec(),
                 Err(_) => {
                     return Err(ApplyError::InternalError(String::from(
```
peterschwarz: Should this also be an InvalidTransaction?

contracts/track_and_trace/src/handler.rs
```diff
@@ -32,136 +29,88 @@ cfg_if! {
...
-impl SupplyChainTransactionHandler {
+impl TrackAndTraceTransactionHandler {
     #[allow(clippy::new_without_default)]
-    pub fn new() -> SupplyChainTransactionHandler {
-        SupplyChainTransactionHandler {
+    pub fn new() -> TrackAndTraceTransactionHandler {
+        TrackAndTraceTransactionHandler {
             family_name: "grid_track_and_trace".to_string(),
             family_versions: vec!["1.0".to_string()],
             namespaces: vec![get_supply_chain_prefix().to_string()],
```
peterschwarz: Need to add both namespaces here.

contracts/track_and_trace/src/state.rs
eloaverona: I guess I was not consistent in either returning InvalidTransaction or InternalError in this deserialization from bytes step. Would it be better to change so that all deserialiation errors are mapped to InternalError? If there is a deserialization error its would be more likely due to a  programmer's mistake and not the user's.

eloaverona: Addressed

contracts/track_and_trace/src/handler.rs
eloaverona: Addressed

contracts/track_and_trace/src/state.rs
```diff
@@ -438,3 +401,301 @@ impl<'a> SupplyChainState<'a> {
...
+    fn test_get_agent_some() {
+        let mut transaction_context = MockTransactionContext::default();
+        transaction_context.add_agent("agent_public_key");
+        let state = TrackAndTraceState::new(&mut transaction_context);
+        let result = state.get_agent("agent_public_key").unwrap();
+        assert_eq!(result, Some(make_agent("agent_public_key")))
+    }
+
+    #[test]
+    // Test that if an record does not exist in state, None is returned
```
dplumb94: Small typo in these comments, should be "a record", etc.

eloaverona: Addressed

PR #82, opened 2019-05-30
Update Track and Trace spec

dplumb94:
To test:

```
$ cd grid
$ docker build . -f docs/grid-build-docs -t grid-build-docs
$ docker run -v $(pwd):/project/grid grid-build-docs
```

Navigate to `grid/docs/build` and open up the html or pdf

### DIFF
--- a/docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
+++ b/docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
@@ -8,22 +8,28 @@ Overview
 The Grid Track and Trace transaction family allows users to track
 goods as they move through a supply chain. Records for goods include a
 history of ownership and custodianship, as well as histories for a
-variety of properties such as temperature and location. These
-properties are managed through a user-specifiable system of record
-types.
+variety of properties such as temperature and location. These properties are
+managed using Grid Schemas.
 
 
 State
 =====
 
 All Grid Track and Trace objects are serialized using Protocol Buffers before
-being stored in state. These objects include: Agents, Properties
-(accompanied by their auxiliary PropertyPage objects), Proposals,
-Records, and RecordTypes. As described in the Addressing_ section
-below, these objects are stored in separate sub-namespaces under the
-Grid Track and Trace namespace. To handle hash collisions, all objects are
-stored in lists within protobuf "Container" objects.
+being stored in state. These objects include: Records, Proposals, and
+Properties (accompanied by their auxiliary PropertyPage objects). As described
+in the Addressing_ section below, these objects are stored in separate
+sub-namespaces under the Grid Track and Trace namespace. To handle hash
+collisions, all objects are stored in lists within protobuf "List" objects.
 
+.. note:: In addition to the messages defined in Grid Track and Trace, this
+    transaction family also makes use of Agents (as defined in the :doc:`Pike
+    Transaction Family<pike_transaction_family>`), as well as Schemas,
+    PropertyDefinitions, and PropertyValues (as defined in the :doc:`Grid
+    Schema Transaction Family<grid_schema_family_specification>`). Clients
+    and contracts that implement this specification will need to orchestrate
+    these transaction families together in order to create a working
+    application.
 
 Records
 -------
@@ -31,7 +37,7 @@ Records
 Records represent the goods being tracked by Grid Track and Trace. Almost
 every transaction references some Record.
 
-A Record contains a unique identifier, the name of a RecordType, and
+A Record contains a unique identifier, the name of a Schema, and
 lists containing the history of its owners and custodians. It also
 contains a ``final`` flag indicating whether further updates can be
 made to the Record and its Properties. If this flag is set to true,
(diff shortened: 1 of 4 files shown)


docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -8,30 +8,35 @@ Overview
 The Grid Track and Trace transaction family allows users to track
 goods as they move through a supply chain. Records for goods include a
 history of ownership and custodianship, as well as histories for a
-variety of properties such as temperature and location. These
-properties are managed through a user-specifiable system of record
-types.
+variety of properties such as temperature and location. These properties are
+managed through a user-specifiable system of schemas.
```
peterschwarz: "managed using Grid Schemas" - it's a specific system.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -40,44 +45,45 @@ its ``final`` flag.
...
+        // changes can be made to it or its Properties.
+        bool final = 5;
+    }
 
 
 Note that while information about a Record's owners and custodians are
 included in the object, information about its Properties are stored
 separately (see the Properties_ section below).
 
 Records whose addresses collide are stored in a list alphabetically by
```
peterschwarz: change this from "a list alphabetically by" to "a list, sorted by"

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -119,206 +125,95 @@ ever runs out of space for new updates. Under this scheme, 16^2 *
...
+
+        // The name of the Property, e.g. "temperature". This must be unique among
+        // Properties.
+        string name = 1;
+
+        // The natural key of the Property's associated Record.
+        string record_id = 2;
+
+        // The name of the PropertyDefinition that defines this record.
+        PropertyDefinition property_definition = 3;
```
peterschwarz: Should you include the protobuf import where this comes from?

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -743,13 +576,17 @@ marked as accepted.
 
 .. code-block:: protobuf
 
-   message RevokeReporterPayload {
```
eloaverona: The explanation of the RevokeReporter action says "The owner of a Record can send a RevokeReporter transaction to remove a reporter's authorization to report on one or more Properties for that Record. This creates a Proposal which is immediately closed and marked as accepted." This last part is not true. A proposal is not created in this transaction and I don't think it should be. There is no proposal for revoking a reporter.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -743,13 +576,17 @@ marked as accepted.
...
-   }
+        // The reporter's public key
+        string reporter_id = 2;
+
+        // The names of the Properties for which the reporter's
+        // authorization is revoked
+        repeated string properties = 3;
+    }
 
 A RevokeReporter transaction is invalid if one of the following
```
eloaverona: A RevokeReporter transaction it is also invalid if any of the provided properties do not exits.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -635,7 +468,7 @@ conditions occurs:
 - The Record is final.
 - Its signer is not authorized to report on that Record.
```
eloaverona: It should be "Its signer is not authorized to report on any of the provided properties". Or something like that. An agent is authorized to report on a particular property in a record, not in the whole record.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -620,12 +453,12 @@ sent by an Agent authorized to report on the Property.
...
+        // The natural key of the Record
+        string record_id = 1;
 
-       repeated PropertyValue properties = 2;
-   }
+        repeated PropertyValue properties = 2;
+    }
 
 
 An UpdateProperties transaction is invalid if one of the following
```
eloaverona: The UpdateProperties transaction also fails if one of the property values provided have a different data_type than the specified in the property definition.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -695,24 +527,25 @@ status as ``canceled`` rather than ``rejected``.
...
+
+        // The role being proposed (owner, custodian, or reporter)
+        Proposal.Role role = 3;
+
+        // The respose to the Proposal (accept, reject, or cancel)
+        Response response = 4;
+    }
 
 
 Proposals can conflict, in the sense that a Record's owner might have
```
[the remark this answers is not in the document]
dplumb94: The last case you mentioned is covered by the last clause in the list. The others are all checking the validity of the record, rather than the validity of the answer. To avoid having a super long list, I think I'll add a clause: "The record is no longer valid". Let me know if you agree with that.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
dplumb94: I'm not sure about that. I checked the Pike spec (which also has imports in the protobufs) and there are no import messages in the code blocks. I didn't want to give the impression that each block is a valid file. The origin of these messages is also called out explicitly in the intro to this section.

However, if you feel strongly about it, I don't disagree with including it.

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -649,37 +482,36 @@ of Property names must be included.
...
 - The signer is not the owner and the Proposal is for transfer of
   ownership or reporter authorization.
 - The signer is not the custodian and the Proposal is for transfer of
   custodianship.
-- The receiving Agent is not registered (the signer must be registered
-  as well, but this is implied by the previous two conditions).
-- There is already an open Proposal for the Record and receiving Agent
-  for the specified role.
-- The Record is final.
 - The Proposal is for reporter authorization and the list of Property
```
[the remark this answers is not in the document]
dplumb94: Added a commit to add this check and a unit test.

sdk/protos/track_and_trace_state.proto
```diff
@@ -65,8 +71,8 @@ message PropertyPage {
   message ReportedValue {
     // The index of the reporter id in reporters field
     uint32 reporter_index = 1;
-    // Approximately when this value was reported, as a Unix UTC
-    // timestamp
+
+    // Approximately this value was reported, as a Unix UTC timestamp.
```
jsmitchell: I think you removed 'when' by mistake. An alternate phrasing would be "The approximate time this value was reported, ..."

docs/source/transaction_family_specifications/grid_track_and_trace_family_specification.rst
```diff
@@ -119,206 +125,95 @@ ever runs out of space for new updates. Under this scheme, 16^2 *
...
 
+    message PropertyPage {
+        message ReportedValue {
+            // The index of the reporter id in reporters field.
+            uint32 reporter_index = 1;
 
-Because it is expected to be used for many RecordTypes, a dedicated
-Location protobuf message is used, the values of which are latitude
-and longitude.
+            // Approximately this value was reported, as a Unix UTC timestamp.
```
jsmitchell: Time at which...

chenette: "Approximately when" is simpler (re the grammatical structure).

How about this?  "When this value was reported (approximately), as ...."

sdk/protos/track_and_trace_state.proto
dplumb94: Whoops.... thanks. I think I like the phrasing you supplied better.

dplumb94: Does it really make sense to say "Approximate"?

PR #83, opened 2019-06-03
Implement GET /record endpoint

dplumb94:
Implements an endpoint that lists all records in the record collection.

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -327,6 +327,32 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /record:
+    get:
+      tags:
+        - Track and Trace
+      summary: Fetch a list of records
+      description: |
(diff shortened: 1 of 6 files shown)


desktophero:
the `diffs` in this PR help illustrate some of the safety aspects of Rust and how to build some other apps in the HL ecosystem. Glad I saw this in the feed. Dig it.

dplumb94:
> the `diffs` in this PR help illustrate some of the safety aspects of Rust and how to build some other apps in the HL ecosystem. Glad I saw this in the feed. Dig it.

Glad this is helpful! Diesel (the query builder used here) is pretty nice, but we've found that it can be tricky to execute highly complex queries. It provides an API for writing raw SQL (`diesel::sql_query`), but doing so sacrifices some of the type safety benefits of Rust. You can also [treat views as tables](https://deterministic.space/diesel-view-table-trick.html), which might get used in a later PR for track and trace integration.

PR #84, opened 2019-06-06
Implement GET /record/{record_id}/{property_name} endpoint

eloaverona:
Implements an endpoint to fetch the information on a particular property of a record, including current value and history of changes.

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -384,6 +384,44 @@ paths:
           $ref: "#/components/responses/500ServerError"
         "503":
           $ref: "#/components/responses/503ServiceUnavailable"
+  "/record/{record_id}/property/{property_name}":
+      get:
+        tags:
+          - Track and Trace
+        summary: Fetch a particular property
+        description: |
+          Fetches a single property with the given record ID and property name.
+        operationId: fetch_property
+        parameters:
+          - name: record_id
+            in: path
+            description: ID of the record to fetch a property from.
+            required: true
+            schema:
+              type: string
+          - name: property_name
+            in: path
+            description: Name of the property to fetch.
+            required: true
+            schema:
+              type: string
+        responses:
+          "200":
+            description: Successful operation
+            content:
+              application/json:
+                schema:
+                  properties:
+                    data:
+                      $ref: "#/components/schemas/Property"
+          "400":
+            $ref: "#/components/responses/400BadRequest"
+          "404":
+            $ref: "#/components/responses/404NotFound"
+          "500":
+            $ref: "#/components/responses/500ServerError"
+          "503":
+            $ref: "#/components/responses/503ServiceUnavailable"
 components:
   parameters:
     batch_id:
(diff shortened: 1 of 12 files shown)


daemon/src/rest_api/routes/records.rs
```diff
@@ -159,3 +168,368 @@ pub fn fetch_record(
...
+                metadata: reported_value_with_reporter
+                    .metadata
+                    .clone()
+                    .unwrap_or_else(|| JsonValue::Object(Map::new())),
+            },
+        })
+    }
+}
+
+fn parse_value(
```
dplumb94: It seems like there must be a Rustier way to do this... maybe @peterschwarz can weight in.

daemon/src/rest_api/routes/records.rs
```diff
@@ -159,3 +168,368 @@ pub fn fetch_record(
...
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct PropertyValueSlice {
+    pub timestamp: u64,
+    pub value: Value,
+    pub reporter: ReporterSlice,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Value(
```
peterschwarz: This should be an enum:

```
pub enum Value {
    String(String),
    Bool(bool),
    Struct(Vec<StructPropertyValue>),
    LatLong(LatLong),
    // ... etc
}
```
where the Serde derive items are specified in the appropriate way as referenced here: https://serde.rs/enum-representations.html#enum-representations

daemon/src/rest_api/routes/records.rs
peterschwarz: See my above comment on `Value`

daemon/src/rest_api/routes/records.rs
```diff
@@ -214,3 +223,369 @@ pub fn fetch_record(
...
+    pub value: Value,
+    pub reporter: ReporterSlice,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub enum Value {
+    String(String),
+    Bool(bool),
+    Struct(Vec<StructPropertyValue>),
+    Location(LatLong),
```
peterschwarz: This should probably be called `LatLong`, as location will probably have a more structured meaning down the road.

daemon/openapi.yaml
```diff
@@ -777,3 +817,66 @@ components:
...
+            - $ref: '#/components/schemas/StructValue'
+            - $ref: '#/components/schemas/LocationValue'
+            - $ref: '#/components/schemas/NumberValue'
+            - $ref: '#/components/schemas/EnumValue'
+            - $ref: '#/components/schemas/BytesValue'
+          example: {String: lighbulb}
+    StringValue:
+      type: object
+      properties:
+        String:
```
dplumb94: This would be a somewhat large change to the API. The JSON return values would look like:

```
"value": {
      "timestamp": 1557949075,
      "value": {
        "Bytes": "AQIDBA=="
      },
      ...
}
```
, where previously it was
```
"value": {
      "timestamp": 1557949075,
      "value":  "AQIDBA==",
}
```

I think the former response is preferable since we already have the data type enum in the property response. I'm not sure what implications that change would have on the UI. Also if it needs to stay this way, that property's key should be snake_case.

eloaverona: Okay, I'll undo this change

PR #87, opened 2019-06-11
Fix bug in Pike contract

eloaverona:
Admin agents should not be able to remove the admin role from
themselves. This commit adds a check and returns an InvalidTransaction
if an admin tries to submit a transaction that would remove the role
admin from themselves.

Signed-off-by: Eloá Franca Verona <eloafran@bitwise.io>

### DIFF
--- a/contracts/pike/src/handler.rs
+++ b/contracts/pike/src/handler.rs
@@ -353,6 +353,17 @@ fn update_agent(
     };
 
     if !payload.get_roles().is_empty() {
+        // verify that an admin is not removing the role admin from themselves.
+        if signer == payload.get_public_key()
+            && !payload.get_roles().iter().any(|role| role == "admin")
+        {
+            return Err(ApplyError::InvalidTransaction(
+                "An admin cannot remove themselves as admin. 'admin' role must be included
+                    in the roles list."
+                    .to_string(),
+            ));
+        }
+
         agent.set_roles(protobuf::RepeatedField::from_vec(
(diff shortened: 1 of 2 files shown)


contracts/pike/src/handler.rs
```diff
@@ -335,6 +335,21 @@ fn update_agent(
     // verify the signer of the transaction is authorized to update agent
     is_admin(signer, payload.get_org_id(), state)?;
 
+    // verify that an admin is not removing the role admin from themselves.
+    if signer == payload.get_public_key()
+        && payload
```
agunde406: This should also check that the roles in the payload are not empty. If the payload does not contain new roles, the roles are not updated. See line 370.

eloaverona: Addressed

contracts/pike/src/handler.rs
```diff
@@ -353,6 +353,21 @@ fn update_agent(
     };
 
     if !payload.get_roles().is_empty() {
+        // verify that an admin is not removing the role admin from themselves.
+        if signer == payload.get_public_key()
+            && payload
+                .get_roles()
+                .iter()
+                .find(|&role| role == "admin")
```
peterschwarz: this `.find(predicate_fn).is_some()` can be replaced with `.any(predicate_fn)`

eloaverona: addressed

PR #89, opened 2019-06-12
Upgrading Actix

knabben:
Moving actix to latest 0.8.3 and actix-web to 1.0.0

Related to: https://jira.hyperledger.org/browse/GRID-78

### DIFF
--- a/daemon/src/rest_api/routes/batches.rs
+++ b/daemon/src/rest_api/routes/batches.rs
@@ -15,7 +15,7 @@
 use crate::rest_api::{error::RestApiResponseError, routes::SawtoothMessageSender, AppState};
 
 use actix::{Context, Handler, Message};
-use actix_web::{AsyncResponder, HttpMessage, HttpRequest, HttpResponse, Query, State};
+use actix_web::{web, HttpRequest, HttpResponse};
 use futures::future;
 use futures::future::Future;
 use sawtooth_sdk::messages::batch::{Batch, BatchList};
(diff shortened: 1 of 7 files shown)


jsmitchell:
Hi @knabben - thanks for the contribution! One of your commits is missing signoff. You can fix that branch by doing `git commit --amend --signoff` and then force pushing.

peterschwarz:
Could you also update your commit messages to follow: https://chris.beams.io/posts/git-commit/  Specifically your second commit gives no context to the change (I thought it was about the DB migrations, and it sounded totally unrelated, for example).

knabben:
sure, after figuring out what is breaking at compilation time, I do both action. Gonna pay more attention on PR commits, in the mean time can some one run this on Jenkins or add my user to the whitelist, so I can see which tests are breaking as well.

daemon/src/rest_api/routes/batches.rs
```diff
@@ -159,9 +159,9 @@ impl Handler<BatchStatuses> for SawtoothMessageSender {
 }
 
 pub fn submit_batches(
-    (req, state): (HttpRequest<AppState>, State<AppState>),
+    (req, state, body): (HttpRequest, web::Data<AppState>, web::Bytes),
 ) -> impl Future<Item = HttpResponse, Error = RestApiResponseError> {
-    req.body().from_err().and_then(
+    body.from_err().and_then(
```
knabben: This is failing with: 
```
error[E0599]: no method named `from_err` found for type `bytes::bytes::Bytes` in the current scope
   --> src/rest_api/routes/batches.rs:164:10
    |
164 |     body.from_err().and_then(
    |          ^^^^^^^^
    |
    = note: the method `from_err` exists but the following trait bounds were not satisfied:
            `&mut [u8] : futures::future::Future`
            `&mut bytes::bytes::Bytes : futures::future::Future`
```

peterschwarz:
Would you mind taking this issue in Jira: https://jira.hyperledger.org/browse/GRID-78

peterschwarz:
Looks like you have lint - run `cargo fmt` and fixup the appropriate commits

daemon/src/rest_api/routes/batches.rs
peterschwarz: it looks like you don't need that `body.from_err().and_then(...`, you can just use the body directly and remove the wrapping function.  Once that is corrected, a whole new host of errors crop up, though.

knabben: True, I need more research on Rust to come up with the final solution, closing this for now. The guide: https://github.com/actix/actix-web/blob/master/MIGRATION.md is a good start.

PR #90, opened 2019-06-12
Move DB Migrations to CLI

eloaverona:
This PR:
- Adds CLI command to run db migrations 
- Removes the capability of running the database migrations when the daemon starts up. 
- Adds a check on daemon start up that the migrations have been applied, and returns an error if they have not.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -16,6 +16,9 @@
 extern crate clap;
 #[macro_use]
 extern crate log;
+#[macro_use]
+extern crate diesel_migrations;
+extern crate diesel;
 
 mod actions;
 mod error;
@@ -35,7 +38,7 @@ use simple_logger;
 
 use crate::error::CliError;
 
-use actions::{agents, keygen, organizations as orgs, schemas};
+use actions::{agents, database, keygen, organizations as orgs, schemas};
 
 const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -109,6 +112,15 @@ fn run() -> Result<(), CliError> {
             (@subcommand show =>
                 (about: "Show schema specified by name argument")
                 (@arg name: +takes_value +required "Name of schema")
+            )
+        )
+        (@subcommand database =>
+            (about: "Manage Grid Daemon database")
+            (@setting SubcommandRequiredElseHelp)
+            (@subcommand migrate =>
+                (about: "Run database migrations")
+                (@arg database_url: --("database-url") +takes_value
+                    "URL for database")
             )
         )
         (@subcommand keygen =>
@@ -204,6 +216,13 @@ fn run() -> Result<(), CliError> {
             }
             ("list", Some(_)) => schemas::do_list_schemas(&url)?,
             ("show", Some(m)) => schemas::do_show_schema(&url, m.value_of("name").unwrap())?,
+            _ => return Err(CliError::UserError("Subcommand not recognized".into())),
+        },
+        ("database", Some(m)) => match m.subcommand() {
+            ("migrate", Some(m)) => database::run_migrations(
+                m.value_of("database_url")
+                    .unwrap_or("postgres://grid:grid_example@localhost/grid"),
+            )?,
             _ => return Err(CliError::UserError("Subcommand not recognized".into())),
         },
         ("keygen", Some(m)) => keygen::generate_keys(
--- a/daemon/Dockerfile
+++ b/daemon/Dockerfile
@@ -55,8 +55,23 @@ RUN cargo build --release
 RUN rm src/*.rs
 COPY ./daemon/src ./src
 COPY ./daemon/migrations ./migrations
+COPY ./daemon/migrations ../daemon/migrations
+
 
 RUN rm ./target/release/gridd* ./target/release/deps/gridd*
+RUN cargo build --release
+
+WORKDIR ..
+RUN USER=root cargo new --bin cli
+WORKDIR /cli
+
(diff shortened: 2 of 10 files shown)


cli/src/main.rs
```diff
@@ -111,6 +114,16 @@ fn run() -> Result<(), CliError> {
                 (@arg name: +takes_value +required "Name of schema")
             )
         )
+        (@subcommand migration =>
+            (about: "Update or create agent")
```
peterschwarz: This looks like a copy-paste error

cli/src/main.rs
```diff
@@ -111,6 +114,16 @@ fn run() -> Result<(), CliError> {
...
+            (@setting SubcommandRequiredElseHelp)
+            (@subcommand run =>
```
peterschwarz: I think the command might be better if it was:

```
$ grid database migrate ...
```

Thoughts, @vaporos ?

daemon/src/error.rs
```diff
@@ -31,6 +31,7 @@ pub enum DaemonError {
     RestApiError(RestApiServerError),
     StartUpError(Box<dyn Error>),
     ShutdownError(String),
+    MigrationError(String),
```
peterschwarz: You could add the list of migrations as part of the error, and then have the message as part of the Display format.

Also, at this level, the error might better be called `DatabaseOutOfDate`

daemon/src/database/mod.rs
```diff
@@ -24,20 +24,33 @@ embed_migrations!("./migrations");
...
 };
 
-pub use super::database::error::DatabaseError;
+use diesel_migrations::mark_migrations_in_directory;
+use std::path::Path;
 
-pub fn run_migrations(database_url: &str) -> Result<(), DatabaseError> {
+/// Checks that all migrations were applied. Returns a list of the migration versions that were
+/// not applied.
+pub fn check_migrations(database_url: &str) -> Result<Vec<String>, DatabaseError> {
```
peterschwarz: I would consider returning `Result<(), DatabaseError>` and have a variant on `DatabaseError` for `MissingMigrations(Vec<String>)`

cli/src/main.rs
eloaverona: Addressed

cli/src/main.rs
eloaverona: Addressed

daemon/src/database/mod.rs
eloaverona: Addressed

daemon/src/error.rs
eloaverona: I added `MissingMigrations(Vec<String>)` variant to `DatabaseError` and refactored the check_migrations function as you suggested.
These changes made it unnecessary to add a new type of error to `DaemonError` as `DaemonError` implements `From<DatabaseError>`.

cli/src/main.rs
```diff
@@ -111,6 +114,16 @@ fn run() -> Result<(), CliError> {
                 (@arg name: +takes_value +required "Name of schema")
             )
         )
+        (@subcommand database =>
+            (about: "Manage Grid Daemon database")
+            (@setting SubcommandRequiredElseHelp)
+            (@subcommand migrate =>
+                (about: "Run database migrations")
+                (@arg path: --path +takes_value "Path to migrations folder")
```
peterschwarz: This does not appear to be used.

daemon/src/database/mod.rs
```diff
@@ -24,18 +24,35 @@ embed_migrations!("./migrations");
...
-pub use super::database::error::DatabaseError;
+use diesel_migrations::mark_migrations_in_directory;
+use std::path::Path;
 
-pub fn run_migrations(database_url: &str) -> Result<(), DatabaseError> {
+/// Checks that all migrations were applied.
+pub fn check_migrations(database_url: &str) -> Result<(), DatabaseError> {
     let connection = PgConnection::establish(database_url)?;
+    let migrations_not_applied =
+        mark_migrations_in_directory(&connection, Path::new("./migrations"))?
```
peterschwarz: Like the CLI, these migrations should be compiled into the binary.  Is there anyway to use the diesel_migrations library to check these in that manner?

eloaverona: The `mark_migrations_in_directory` method is from the `diesel_migrations` library.  But I did not find a way to do something similar for the embed migrations.

cli/src/main.rs
eloaverona: I updated the PR to remove that argument.

daemon/src/database/mod.rs
eloaverona: I'll take a deeper look into it.

daemon/Dockerfile
```diff
@@ -68,5 +83,7 @@ RUN apt-get update \
  && rm -rf /var/lib/apt/lists/*
 
 COPY --from=BUILDER /daemon/target/release/gridd /
+COPY --from=BUILDER /cli/target/release/grid /
+
 
-CMD ["/gridd"]
+CMD ["/gridd", "/grid"]
```
peterschwarz: I think these end up as arguments - I wouldn't make this particular change.

eloaverona: Addressed

daemon/src/database/mod.rs
```diff
@@ -25,21 +25,12 @@ embed_migrations!("./migrations");
 use std::ops::Deref;
 
 use diesel::{
-    connection::Connection as _,
```
agunde406: Typo in commit message: starup ->  startup

eloaverona: Updated

PR #92, opened 2019-06-14
Add keygen command to CLI

eloaverona:
This PR adds the keygen command to the CLI, this command generates a public/private key pair that can be used to sign transactions.

Signed-off-by: Eloá Franca Verona <eloafran@bitwise.io>

### DIFF
--- a/cli/src/actions/keygen.rs
+++ b/cli/src/actions/keygen.rs
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2018 Intel Corporation
+ * Copyright 2019 Cargill Incorporated
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


cli/src/actions/keygen.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::fs::{self, OpenOptions};
+use std::io::prelude::*;
+use std::os::unix::fs::OpenOptionsExt;
+use std::path::PathBuf;
+use users::get_current_username;
+
+/// Generates a public/private key pair that can be used to sign transactions.
+/// If no directory is provided, the keys are created in the default directory
+///
+///   $HOME/.sawtooth/keys/
```
peterschwarz: Should this be `.grid`?

eloaverona: I guess that would be better. Though, `$HOME/.sawtooth/keys/` is the directory that the `key` module in the CLI looks for keys by default . I can change that to `$HOME/.grid/keys/` as well.

eloaverona: Addressed

PR #96, opened 2019-06-19
Make contracts dir before cargo new

dplumb94:
The updated version of cargo in the nightly toolchain throws an error on `cargo new --bin contracts/[CONTRACT]` if the contracts directory doesn't exist.

Signed-off-by: Darian Plumb <dplumb@bitwise.io>

### DIFF
--- a/docker/tests
+++ b/docker/tests
@@ -59,19 +59,19 @@ COPY ./contracts/schema/Cargo.toml ./Cargo.toml
 RUN cargo check
 
 WORKDIR /
-RUN USER=root cargo new --bin daemon
+RUN USER=root cargo new --bin daemon --vcs none
 WORKDIR /daemon
 COPY ./daemon/Cargo.toml ./Cargo.toml
 RUN cargo check
 
 WORKDIR /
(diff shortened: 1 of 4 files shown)


daemon/openapi.yaml
```diff
@@ -710,6 +710,9 @@ components:
         record_id:
           type: string
```
eloaverona: Should this commit be included in this PR?

dplumb94: fixt

docker/tests
```diff
@@ -50,27 +50,28 @@ RUN rustup update \
 
 COPY ./sdk /sdk
 
-RUN USER=root cargo new --bin contracts/schema
+RUN mkdir contracts \
+ && USER=root cargo new --bin contracts/schema
```
eloaverona: Should you add the ` --vcs none` flag here too? And perhaps to the other Dockerfiles modified in this PR?

dplumb94: For some reason this does not cause an error. It may just be an issue with nightly cargo.

PR #97, opened 2019-06-20
Update property update return value

dnewh:
Update the method that fetched property update values so that it
returns an array that only contains one copy of each update. This
simplifies the payload returned in API calls to the /records/:recordId?
endpoints.

### DIFF
--- a/daemon/src/database/helpers/track_and_trace.rs
+++ b/daemon/src/database/helpers/track_and_trace.rs
@@ -363,7 +363,7 @@ pub fn list_reported_value_reporter_to_agent_metadata(
                 .and(reported_value_reporter_to_agent_metadata::record_id.eq(record_id))
                 .and(
                     reported_value_reporter_to_agent_metadata::reported_value_end_block_num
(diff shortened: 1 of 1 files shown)


daemon/src/database/helpers/track_and_trace.rs
```diff
@@ -363,7 +363,7 @@ pub fn list_reported_value_reporter_to_agent_metadata(
                 .and(reported_value_reporter_to_agent_metadata::record_id.eq(record_id))
                 .and(
                     reported_value_reporter_to_agent_metadata::reported_value_end_block_num
-                        .le(MAX_BLOCK_NUM),
```
eloaverona: I don't think this update will fix the problem you are trying to fix. [This method is used](https://github.com/hyperledger/grid/blob/2c56a0dd6225d1e82251b59848f1e7f6610965ed/daemon/src/rest_api/routes/records.rs#L520) to fetch the history of updates for a particular property. Getting only the values that have the end_block_num == MAX_BLOCK_NUM, would not allow for that.

PR #103, opened 2019-06-26
Update Grid to use Sabre 0.4 and Add logging to Smart Contracts

agunde406:
You may need to blow away your local version of the Sabre TP to test.

### DIFF
--- a/cli/src/actions/schemas.rs
+++ b/cli/src/actions/schemas.rs
@@ -15,7 +15,8 @@
 use crate::http::submit_batches;
 use crate::transaction::{schema_batch_builder, GRID_SCHEMA_NAMESPACE, PIKE_NAMESPACE};
 use grid_sdk::protocol::schema::payload::{
-    Action, SchemaCreateBuilder, SchemaPayload, SchemaPayloadBuilder, SchemaUpdateBuilder,
+    Action, SchemaCreateAction, SchemaCreateBuilder, SchemaPayload, SchemaPayloadBuilder,
+    SchemaUpdateAction, SchemaUpdateBuilder,
 };
 use grid_sdk::protocol::schema::state::{DataType, PropertyDefinition, PropertyDefinitionBuilder};
 use grid_sdk::protos::IntoProto;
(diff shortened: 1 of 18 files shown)


sdk/src/protocol/schema/payload.rs
```diff
@@ -26,79 +26,58 @@ use crate::protos::{
...
+        let action = match payload.get_action() {
+            protos::schema_payload::SchemaPayload_Action::SCHEMA_CREATE => Action::SchemaCreate(
+                SchemaCreateAction::from_proto(payload.get_schema_create().clone())?,
+            ),
+            protos::schema_payload::SchemaPayload_Action::SCHEMA_UPDATE => Action::SchemaUpdate(
+                SchemaUpdateAction::from_proto(payload.get_schema_update().clone())?,
+            ),
+            protos::schema_payload::SchemaPayload_Action::UNSET_ACTION => {
+                return Err(ProtoConversionError::InvalidTypeError(
+                    "Cannot convert SchemaPayload_Action with type unset.".to_string(),
```
vaporos: Error strings shouldn't end in a period.

agunde406: Added a commit to remove the period from this error and the other instances.

cli/src/actions/schemas.rs
```diff
@@ -206,7 +206,7 @@ fn generate_create_schema_payload(
         CliError::PayloadError(format!("Failed to build schema payload: {}", err))
     })?;
 
-    schema_paylod = schema_paylod.with_schema_create(schema_create_action);
+    schema_paylod = schema_paylod.with_action(Action::SchemaCreate(schema_create_action));
```
vaporos: Typo in the existing code - schema_paylod

cli/src/actions/schemas.rs
```diff
@@ -328,7 +325,7 @@ fn parse_value_as_sequence(property: &Mapping, key: &str) -> Result<Option<Seque
         Some(value) => match value.as_sequence() {
             Some(value) => Ok(Some(value.to_vec())),
             None => Err(CliError::InvalidYamlError(format!(
-                "Value of {} has an invalid format. Expected is a yaml list.",
+                "Value of {} has an invalid format. Expected is a yaml list",
```
vaporos: These would be better without the internal period. "Expected is a" is awkward.

In other messages, we have "`" characters, which if they make it to the command line, have a meaning in bash. Better to avoid those special characters in error messages.

There are all existing though, so I don't think we should necessarily fix them in this PR.

agunde406: I am going to remove the period removal commit and open up another PR and move this comment.

PR #104, opened 2019-06-26
Add product db schemas/models

pa3ng:
Adding grid product db queries, schemas, models, and helper functions.

cc: @adeebahmed

### DIFF
--- a/daemon/src/database/helpers/products.rs
+++ b/daemon/src/database/helpers/products.rs
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2019 Target Brands, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


daemon/src/database/helpers/products.rs
```diff
@@ -0,0 +1,78 @@
...
+    pg::PgConnection,
+    prelude::*,
+    result::Error::NotFound,
+    QueryResult,
+};
+
+#[allow(dead_code)]
+pub fn insert_products(conn: &PgConnection, products: &[NewProduct]) -> QueryResult<()> {
+    for prod in products {
+        update_prod_end_block_num(conn, &prod.prod_id, prod.start_block_num)?;
```
peterschwarz: These should be updated to the current block num, not the start block. Otherwise, there will be gaps in history where the product is "missing".

pa3ng: @peterschwarz Maybe I'm misunderstanding something here, but isn't `prod.start_block_num` being treated as the current block number in  `fn update_prod_end_block_num`? And then updated here https://github.com/hyperledger/grid/blob/c45a186d49a8a1f5240bd6c5fce9b7959bc2a1ea/daemon/src/database/helpers/products.rs#L53 (as with the other update functions for schemas, organizations, etc.)

pa3ng:
@peterschwarz I generated the database migrations directory for product via `diesel migration generate create_product_table`

PR #108, opened 2019-07-24
Add product properties database field

pa3ng:
Adding the `properties` field to the Product database schema/model (replaces `metatdata` field)


daemon/migrations/2019-07-03-161310_create_product_table/up.sql
```diff
@@ -18,7 +18,7 @@ CREATE TABLE IF NOT EXISTS product (
     prod_id VARCHAR(256) NOT NULL,
     prod_type TEXT NOT NULL,
     owner VARCHAR(256) NOT NULL,
-    metadata JSON [] NOT NULL
+    properties TEXT []
```
vaporos: Should this be in a new migration?

pa3ng: Not exactly sure. @peterschwarz any thoughts?

pa3ng:
Closing PR. Will include database field updates in the next PR for updating grid state delta export

PR #113, opened 2019-08-06
Add Grid product to state delta export

pa3ng:
Updating the state delta subscriber to handle grid product events.

See [Jira story](https://jira.hyperledger.org/browse/GRID-21)

### DIFF
--- a/daemon/migrations/2019-07-03-161310_create_product_table/up.sql
+++ b/daemon/migrations/2019-07-03-161310_create_product_table/up.sql
@@ -15,11 +15,10 @@
 
 CREATE TABLE IF NOT EXISTS product (
     id BIGSERIAL PRIMARY KEY,
-    prod_id VARCHAR(256) NOT NULL,
-    prod_type TEXT NOT NULL,
-    owner VARCHAR(256) NOT NULL,
-    metadata JSON [] NOT NULL
+    product_id VARCHAR(256) NOT NULL,
+    product_namespace TEXT NOT NULL,
+    owner VARCHAR(256) NOT NULL
 ) INHERITS (chain_record);
 
-CREATE INDEX IF NOT EXISTS prod_id_block_num_idx
-    ON product (prod_id, end_block_num);
+CREATE INDEX IF NOT EXISTS product_idx
+    ON product (product_id, end_block_num);
--- a/daemon/src/event/block.rs
+++ b/daemon/src/event/block.rs
@@ -20,7 +20,8 @@ use diesel::result::Error;
 use grid_sdk::{
     protocol::{
         pike::state::{AgentList, OrganizationList},
-        schema::state::{DataType, PropertyDefinition, SchemaList},
+        product::state::ProductList,
+        schema::state::{DataType, PropertyDefinition, PropertyValue, SchemaList},
         track_and_trace::state::{
             PropertyList, PropertyPageList, ProposalList, RecordList, ReportedValue,
         },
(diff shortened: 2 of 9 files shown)


daemon/migrations/2019-07-03-161310_create_product_table/up.sql
```diff
@@ -22,3 +22,20 @@ CREATE TABLE IF NOT EXISTS product (
 
 CREATE INDEX IF NOT EXISTS product_idx
     ON product (product_id);
+
+CREATE TABLE IF NOT EXISTS product_property_value (
```
peterschwarz: Should this be in another migration?

daemon/src/event/block.rs
```diff
@@ -397,6 +398,40 @@ fn state_change_to_db_operation(
...
+                })
+                .collect::<Vec<(NewProduct, Vec<NewProductPropertyValue>)>>();
+
+            let properties = product_list
+                .clone()
+                .into_iter()
+                .flat_map(|(_, pv)| pv.into_iter())
+                .collect();
+
+            let products = product_list.into_iter().map(|(p, _)| p).collect();
```
peterschwarz: This could possibly avoid the clone by using [fold](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.fold) and a tuple of vectors

peterschwarz: This could be done with [fold](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.fold) and a tuple of vecs to put the results.

daemon/migrations/2019-07-03-161310_create_product_table/up.sql
pa3ng: @peterschwarz It could be; I was simply following the setup for grid schema tables: https://github.com/hyperledger/grid/blob/master/daemon/migrations/2019-04-19-191743_create_grid_schema_tables/up.sql#L26 I can move it to another migration if that makes more sense

pa3ng: ✅  (see migration timestamp)

daemon/src/event/block.rs
pa3ng: ✅

pa3ng: ✅

daemon/migrations/2019-07-03-161310_create_product_table/up.sql
pa3ng: ✅ added migration `2019-08-08-145850_create_product_property_table`

daemon/migrations/2019-07-03-161310_create_product_table/up.sql
```diff
@@ -15,11 +15,10 @@
...
-    owner VARCHAR(256) NOT NULL,
-    metadata JSON [] NOT NULL
+    product_id VARCHAR(256) NOT NULL,
+    product_namespace TEXT NOT NULL,
+    owner VARCHAR(256) NOT NULL
 ) INHERITS (chain_record);
 
-CREATE INDEX IF NOT EXISTS prod_id_block_num_idx
-    ON product (prod_id, end_block_num);
+CREATE INDEX IF NOT EXISTS product_idx
```
peterschwarz: What's the reasoning for changing this index?

pa3ng: @peterschwarz Simply to follow the tnt table naming scheme, e.g. `record_idx`, `proposal_idx`, etc.  I can revert it back if you think the change is unnecessary.

peterschwarz: The end block part of the index important to help sort those elements in the index, and it's a common part of the query.

PR #115, opened 2019-08-13
Implement Product REST API

trevormcdonald:
see [Jira](https://jira.hyperledger.org/browse/GRID-23).


agunde406:
Instead of including merge commits, can you please rebase on master and force push

daemon/src/rest_api/routes/products.rs
```diff
@@ -0,0 +1,193 @@
+// Copyright 2019 Cargill Incorporated
```
agunde406: This should be // Copyright (c) 2019 Target Brands, Inc.

PR #116, opened 2019-08-16
Reduce GS1 Product address to 70 bytes

Patrick-Erichsen:
The length of the GS1 product address was 80 bytes.
This commit reduces the product_id hash to 60 bytes
to account for the 10 bytes taken up by the GS1
Product prefix.

Also adds clarity by calculating the length based on
the prefix length rather than using a magic number.

Signed-off-by: Patrick-Erichsen <Patrick.Erichsen@target.com>


contracts/product/src/addressing.rs
```diff
@@ -35,7 +36,10 @@ pub fn hash(to_hash: &str, num: usize) -> String {
 }
 
 pub fn make_product_address(product_id: &str) -> String {
-    get_product_prefix() + PRODUCT + STANDARD + &hash(product_id, 70)
+    let grid_product_gs1_prefix = get_product_prefix() + PRODUCT_NAMESPACE + GS1_NAMESPACE;
+    let grid_product_gs1_prefix_len = grid_product_gs1_prefix.chars().count();
+
+    grid_product_gs1_prefix + &hash(product_id, GRID_ADDRESS_LEN - grid_product_gs1_prefix_len)
```
peterschwarz: Seems like you could just do the math ahead of time, given that this could be a frequently used calculation.

Patrick-Erichsen: Good call - updated this to `let hash_len = GRID_ADDRESS_LEN - grid_product_gs1_prefix_len;`

PR #119, opened 2019-08-21
Remove constraint on ProductList length

Patrick-Erichsen:
When deleting a Product, it is likely that there is only a single Product at the given address. This is not possible with the current `is_empty` constraint on the ProductListBuilder.

This commit removes that constraint to allow for Product deletion.

Signed-off-by: Patrick-Erichsen <Patrick.Erichsen@target.com>

### DIFF
--- a/sdk/src/protocol/product/state.rs
+++ b/sdk/src/protocol/product/state.rs
@@ -367,16 +367,6 @@ impl ProductListBuilder {
             ProductListBuildError::MissingField("'products' field is required".to_string())
         })?;
 
-        let products = {
-            if products.is_empty() {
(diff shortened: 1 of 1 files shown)


sdk/src/protocol/product/state.rs
```diff
@@ -367,16 +367,6 @@ impl ProductListBuilder {
             ProductListBuildError::MissingField("'products' field is required".to_string())
         })?;
 
-        let products = {
-            if products.is_empty() {
```
peterschwarz: This is fixing the wrong problem: on deletion, before constructing this list, the contract should check for empty and, if so, delete the whole state entry, not set an empty list.

PR #121, opened 2019-08-27
Implement Product Database Operations

trevormcdonald:
- Detect StateChange_Type::DELETE in state delta subscriber for GRID_PRODUCT
- Add RemoveProduct(address, current_block_num) as a DbInsertOperation
- Add delete_product and delete_product_property_values functions to db helpers
- Add product_address field to product and product_property_value tables
- Add product_address field to associated Product and ProductPropertyValue structs
- Update all product properties by property_id instead of property_name

See [Jira](https://jira.hyperledger.org/browse/GRID-105)

### DIFF
--- a/daemon/src/rest_api/routes/products.rs
+++ b/daemon/src/rest_api/routes/products.rs
@@ -31,6 +31,7 @@ use std::collections::HashMap;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ProductSlice {
     pub product_id: String,
+    pub product_address: String,
     pub product_namespace: String,
     pub owner: String,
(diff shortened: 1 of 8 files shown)


daemon/src/rest_api/routes/products.rs
```diff
@@ -31,6 +31,7 @@ use std::collections::HashMap;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ProductSlice {
     pub product_id: String,
+    pub product_address: String,
```
agunde406: This looks like a different fix then what the commit message says. Please either change the commit message or split out into different commits.

PR #114, opened 2019-08-09
Implement Product CLI

Patrick-Erichsen:
This PR implements the CLI for Grid Product. 

Actions included are:

- Create
- Update
- Delete
- List
- Show

PR #122, opened 2019-09-03
Add integration testing for Product

trevormcdonald:
- Run tests with `docker-compose -f integration/docker-compose.yaml up`
- Create tests that call CLI functions and check their effects
- Test product create success
- Test product update success
- Test product delete success
- Add these tests to the `bin/run_integration_tests` script
- Add calls for CLI unit tests in `daemon/test/docker-compose.yaml` and `docker/tests`
Note: the unit tests are called by name to avoid also calling integration tests, which
will fail without all services running.
- Format docker compose yaml

This adds a separate docker-compose.yaml (`integration/docker-compose.yaml`) to run integration tests. The compose file is the same as the default `docker-compose.yaml` with logging turned down and a `cargo test` command to run the integration tests after starting the daemon. An optional `INTEGRATION_TEST_URL` can be provided, the default being `http://gridd:8080`.

The tests live in `cli/tests/product.rs`, according to Rust convention, and use the [assert_cmd](https://github.com/assert-rs/assert_cmd) crate recommended by [clap](https://github.com/clap-rs/clap#related-crates). `assert_cmd` takes the rust binary and runs it as a process with given args. The tests currently only verify that the commands do not return errors.

This also adds calls for unit tests in `daemon/test/docker-compose.yaml` and `docker/tests`.

See [Jira](https://jira.hyperledger.org/browse/GRID-106)
Depends on [PR 114](https://github.com/hyperledger/grid/pull/114) and [PR 115](https://github.com/hyperledger/grid/pull/115)

### DIFF
--- a/daemon/test/docker-compose.yaml
+++ b/daemon/test/docker-compose.yaml
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ------------------------------------------------------------------------------
-
-version: '2.1'
+version: "2.1"
 
 services:
     test_server:
@@ -26,24 +25,27 @@ services:
             POSTGRES_PASSWORD: grid_test
             POSTGRES_DB: grid_test
 
-    daemon:
-        build:
-          context: ../..
-          dockerfile: docker/tests
-          args:
-            - http_proxy
-            - https_proxy
-            - no_proxy
-        image: grid:tests
-        volumes:
-          - ../..:/project/grid
-        command: |
-          bash -c "
-            # we need to wait for the db to have started.
-            until PGPASSWORD=grid_test psql -h test_server -U grid_test -c '\q'; do
-                >&2 echo \"Database is unavailable - sleeping\"
-                sleep 1
-            done
-            cd daemon
-            cargo test --features test-api --  --test-threads=1
-            "
+  daemon:
+    build:
+      context: ../..
+      dockerfile: docker/tests
+      args:
+        - http_proxy
+        - https_proxy
+        - no_proxy
+    image: grid:tests
+    volumes:
+      - ../..:/project/grid
+    command: |
+      bash -c "
+        # we need to wait for the db to have started.
+        until PGPASSWORD=grid_test psql -h test_server -U grid_test -c '\q'; do
+            >&2 echo \"Database is unavailable - sleeping\"
+            sleep 1
+        done
+        cd daemon
+        cargo test --features test-api --  --test-threads=1
+        cd ../cli
+        cargo test actions -- --test-threads=1
+        cargo test yaml_parser -- --test-threads=1
+        "
--- a/daemon/test/docker-compose.yaml
+++ b/daemon/test/docker-compose.yaml
@@ -15,15 +15,15 @@
 version: "2.1"
 
 services:
-    test_server:
-        image: postgres
-        restart: always
-        ports:
-            - '5433:5432'
-        environment:
-            POSTGRES_USER: grid_test
-            POSTGRES_PASSWORD: grid_test
-            POSTGRES_DB: grid_test
+  test_server:
+    image: postgres
+    restart: always
+    ports:
+      - "5433:5432"
+    environment:
+      POSTGRES_USER: grid_test
+      POSTGRES_PASSWORD: grid_test
+      POSTGRES_DB: grid_test
 
   daemon:
     build:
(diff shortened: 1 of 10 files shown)


cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,86 @@
+extern crate assert_cmd;
```
agunde406: missing license

cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+    static ORG_NAME: &str = "target";
+    static ORG_ADDRESS: &str = "target hq";
+
+    static PRODUCT_FILE: &str = "/build/cli/tests/test_product.yaml";
+
+    mod product {
+        use super::*;
+        #[test]
+        #[ignore]
+        fn test_product_create() {
```
agunde406: Please add description to this test

cli/tests/integration_tests.rs
```diff
@@ -0,0 +1,86 @@
...
+    static ORG_ID: &str = "314156";
+    static ORG_NAME: &str = "target";
+    static ORG_ADDRESS: &str = "target hq";
+
+    static PRODUCT_FILE: &str = "/build/cli/tests/test_product.yaml";
+
+    mod product {
+        use super::*;
+        #[test]
+        #[ignore]
```
agunde406: Why is this ignored?

cli/tests/integration_tests.rs
trevormcdonald: 👍

cli/tests/integration_tests.rs
trevormcdonald: 👍

cli/tests/integration_tests.rs
trevormcdonald: 👍

daemon/test/docker-compose.yaml
```diff
@@ -1,3 +1,5 @@
+--- # ------------------------------------------------------------------------------
```
vaporos: Was this intentional?

daemon/test/docker-compose.yaml
```diff
@@ -11,39 +13,40 @@
...
-    test_server:
-        image: postgres
-        restart: always
-        ports:
-            - '5433:5432'
-        environment:
-            POSTGRES_USER: grid_test
-            POSTGRES_PASSWORD: grid_test
-            POSTGRES_DB: grid_test
+  test_server:
```
vaporos: Please move the file reformatting to a separate commit (it can still be in the same PR), so it is more obvious what the changes are vs. just reformatting. (Also provide justification for the reformatting in the commit message.)

daemon/test/docker-compose.yaml
```diff
@@ -11,9 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# ------------------------------------------------------------------------------
```
agunde406: This line should not be removed in this commit. You fix it in the next commit.

daemon/test/docker-compose.yaml
```diff
@@ -26,24 +24,27 @@ services:
...
+      bash -c "
+        # we need to wait for the db to have started.
+        until PGPASSWORD=grid_test psql -h test_server -U grid_test -c '\q'; do
+            >&2 echo \"Database is unavailable - sleeping\"
+            sleep 1
+        done
+        cd daemon
+        cargo test --features test-api --  --test-threads=1
+        cd ../cli
+        cargo test actions -- --test-threads=1
```
agunde406: Can you explain the additions of ` cargo test actions -- --test-threads=1` and ` cargo test yaml_parser -- --test-threads=1` in a commit message? Should this a part of the malformed commit?

integration/docker-compose.yaml
```diff
@@ -0,0 +1,269 @@
+--- # ------------------------------------------------------------------------------
```
agunde406: You have an extra --- on the beginning of the line.

daemon/test/docker-compose.yaml
trevormcdonald: I update the first commit message to include:

- Add these tests to the `bin/run_integration_tests` script
- Add calls for CLI unit tests in `daemon/test/docker-compose.yaml` and `docker/tests`
Note: the unit tests are called by name to avoid also calling integration tests, which
will fail without all services running.

I couldn't seem to find a better way to exclude the new tests with `cargo`.

PR #130, opened 2019-12-09
Splinter Mode

RyanLassigBanks:
Adds BatchSubmitter trait to replace Actix handlers for submitting batches, and refactors submit_batch and get_batch_status routes to use it

### DIFF
--- a/daemon/src/rest_api/mod.rs
+++ b/daemon/src/rest_api/mod.rs
@@ -20,19 +20,35 @@ use std::thread;
 
 use crate::database::ConnectionPool;
 pub use crate::rest_api::error::RestApiServerError;
+use crate::rest_api::routes::DbExecutor;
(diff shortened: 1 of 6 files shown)


daemon/src/rest_api/mod.rs
```diff
@@ -20,21 +20,37 @@ use std::thread;
...
     database_connection: Addr<DbExecutor>,
 }
 
+impl AppState {
+    pub fn new(
+        batch_submitter: Box<dyn BatchSubmitter + 'static>,
+        connection_pool: ConnectionPool,
+    ) -> Self {
+        let database_connection =
+            SyncArbiter::start(2, move || DbExecutor::new(connection_pool.clone()));
```
jsmitchell: consider declaring this num threads arg a named constant for readability

PR #135, opened 2019-12-10
Move sawtooth batch submitter

RyanLassigBanks:
Moves sawtooth batch submitter into sawtooth module and moves batch submitter trait into a new batches submodule


daemon/src/rest_api/mod.rs
```diff
@@ -12,29 +12,45 @@
...
     database_connection: Addr<DbExecutor>,
 }
 
+impl AppState {
+    pub fn new(
+        batch_submitter: Box<dyn BatchSubmitter + 'static>,
+        connection_pool: ConnectionPool,
+    ) -> Self {
+        let database_connection =
+            SyncArbiter::start(2, move || DbExecutor::new(connection_pool.clone()));
```
jsmitchell: wouldn't hurt to make this num threads arg a named constant declared above

daemon/src/main.rs
```diff
@@ -26,6 +26,7 @@ extern crate log;
 #[macro_use]
 extern crate serde_json;
 
+mod batches;
```
vaporos: Most module names tend to be singular. Really this is less about 'Batch' as a data structure and more about interfacing with the validator or scabbard, so I wonder if 'submitter' as a module would work better.

RyanLassigBanks: Changed name to submitter.rs

PR #140, opened 2019-12-11
Add prefixes to connect endpoint

RyanLassigBanks:
Adds two type of prefixes that can be added to the `--connect` url.
`--connect` now accepts the following optional prefixes:

`sawtooth` - used to specfiy that the url is for connecting to a
sawtooth validator.

`splinter` - used to specfiy that the url is for connecting to a
splinter node.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -80,7 +80,10 @@ fn run() -> Result<(), DaemonError> {
     let sawtooth_connection = if config.endpoint().is_sawtooth() {
         SawtoothConnection::new(&config.endpoint().url())
     } else {
-        panic!("Unsupported endpoint type: {}", config.endpoint().url());
+        return Err(DaemonError::UnsupportedEndpoint(format!(
+            "Unsupported endpoint type: {}",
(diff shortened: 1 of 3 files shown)


jsmitchell:
Is the daemon meant to panic if a splinter url is provided?

RyanLassigBanks:
> Is the daemon meant to panic if a splinter url is provided?

Currently it should panic because we don't have the splinter backed `BatchSubmitter` implemented yet

daemon/src/main.rs
```diff
@@ -77,7 +77,11 @@ fn run() -> Result<(), DaemonError> {
         .with_cli_args(&matches)
         .build()?;
 
-    let sawtooth_connection = SawtoothConnection::new(config.validator_endpoint());
+    let sawtooth_connection = if config.endpoint().is_sawtooth() {
+        SawtoothConnection::new(&config.endpoint().url())
+    } else {
+        panic!("Unsupported endpoint type: {}", config.endpoint().url());
```
vaporos: This should produce a DaemonError instead.

PR #148, opened 2020-01-07
Adds source query parameter

RyanLassigBanks:
Adds `service_id` query parameter to agent, organization, schema, product, and record routes. If grid is connected to splinter a `source` needs to be added to all queries, specifying a circuit id 

```
http://localhost:8080/agents?service_id="circuit_1234"
```


daemon/src/database/helpers/agents.rs
```diff
@@ -54,21 +54,39 @@ fn update_agent_end_commit_num(
         .map(|_| ())
 }
 
-pub fn get_agents(conn: &PgConnection) -> QueryResult<Vec<Agent>> {
-    agent::table
+pub fn get_agents(conn: &PgConnection, source: &Option<String>) -> QueryResult<Vec<Agent>> {
```
peterschwarz: The source option might make more sense to be `Option<&str>`, and the call sight can pass in `source_opt.as_ref()`, instead of `&source_opt`

daemon/src/database/helpers/agents.rs
```diff
@@ -54,21 +54,39 @@ fn update_agent_end_commit_num(
...
+    let mut query = agent::table
+        .into_boxed()
         .select(agent::all_columns)
-        .filter(agent::end_commit_num.eq(MAX_COMMIT_NUM))
-        .load::<Agent>(conn)
+        .filter(agent::end_commit_num.eq(MAX_COMMIT_NUM));
+
+    if let Some(source) = source {
```
peterschwarz: Does passing an the option to `eq` not work?

If not, it might be interesting to implement something using a trait for all `AsQuery` to add this to the query dsl.  See https://diesel.rs/guides/extending-diesel/

daemon/src/rest_api/routes/mod.rs
```diff
@@ -47,6 +47,7 @@ impl DbExecutor {
 #[cfg(all(feature = "test-api", test))]
 mod test {
     use super::*;
+    use crate::config::Endpoint;
```
peterschwarz: This change looks like it should be in the endpoint commit.

daemon/src/database/helpers/agents.rs
RyanLassigBanks: It does, but I don't think the behavior is desired. If the value is `None` the sql query generated is `source = NULL` which is always false because NULL != NULL is sql.

https://github.com/diesel-rs/diesel/issues/1306

peterschwarz: That issue does have some nice macro suggestions for cleaning that up.

daemon/src/rest_api/routes/mod.rs
RyanLassigBanks: Updated

PR #149, opened 2020-01-07
Implement Grid App Auth Handler

dplumb94:
Adds an initial implementation of an Application Authorization Handler for Splinter to Grid. At runtime, gridd will determine which mode to run by checking the connection string prefix. If the corresponding feature is enabled, Grid will run in that mode.

For example, starting gridd with the command `$ gridd -C splinter:http://splinterd:8085` would require the `splinter-support` feature to be enabled.

Also updated `docker-compose-splinter.yaml` to connect to run Grid in Splinter mode and connect to the splinterd docker-compose service.


daemon/src/main.rs
```diff
@@ -81,72 +86,108 @@ fn run() -> Result<(), DaemonError> {
         .with_cli_args(&matches)
         .build()?;
 
-    let sawtooth_connection = if config.endpoint().is_sawtooth() {
-        SawtoothConnection::new(&config.endpoint().url())
+    let connection_pool = database::create_connection_pool(config.database_url())?;
+
+    if cfg!(feature = "sawtooth-support") && config.endpoint().is_sawtooth() {
+        #[cfg(feature = "sawtooth-support")]
```
RyanLassigBanks: I'm not sure if this needs to be in a feature block. My understanding was that gridd should be compiled to support both splinter and sawtooth and check the "mode" at runtime

daemon/src/config.rs
```diff
@@ -59,8 +63,8 @@ impl From<&str> for Endpoint {
 
         if s.starts_with("splinter:") {
             let mut url = s.replace("splinter:", "");
-            if !url.starts_with("tcp://") {
-                url.insert_str(0, "tcp://");
+            if !url.starts_with("http://") {
```
RyanLassigBanks: [This Pr](https://github.com/hyperledger/grid/pull/151) removes all the protocol checks. This might not be necessary for sawtooth, which is always tcp, but I believe splinter will eventually support multiple protocols, so http and tcp could be valid for splinter mode

daemon/src/main.rs
peterschwarz: This guard as at the wrong level, since that `is_sawtooth` function should presumably be behind the same feature. Same can be said for the splinter block of code.

The contents of these blocks should probably be in functions, as well.

daemon/src/config.rs
```diff
@@ -51,6 +51,10 @@ impl Endpoint {
     pub fn is_sawtooth(&self) -> bool {
         self.backend == Backend::Sawtooth
     }
+
```
peterschwarz: Should this also be guarded by the splinter-support feature?

daemon/src/main.rs
```diff
@@ -150,6 +162,59 @@ fn run() -> Result<(), DaemonError> {
     Ok(())
 }
 
+#[cfg(not(feature = "sawtooth-support"))]
+fn run_sawtooth(config: GridConfig, _connection_pool: ConnectionPool) -> Result<(), DaemonError> {
+    return Err(DaemonError::UnsupportedEndpoint(format!(
+        "A Sawtooth connection endpoint ({}) was provided but Sawtooth support is not enabled.",
```
peterschwarz: Suggest "not enabled in this binary".  The wording suggests that the user could enable it somehow.

daemon/src/config.rs
dplumb94: I don't think so, because the `run()` function checks what mode the given endpoint requires first, and then returns an error if that endpoint is not supported.

PR #151, opened 2020-01-08
Remove prepending of protocol

RyanLassigBanks:
Endpoint not long prepends a protocol to the string that's passed to it,
and makes no assertions about the protocol type

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -55,31 +55,21 @@ impl Endpoint {
 
 impl From<&str> for Endpoint {
     fn from(s: &str) -> Self {
-        let mut s = s.to_lowercase();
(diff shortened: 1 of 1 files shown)


daemon/src/config.rs
```diff
@@ -248,7 +238,7 @@ mod test {
             endpoint,
             Endpoint {
                 backend: Backend::Splinter,
-                url: "tcp://localhost:8080".into()
+                url: "localhost:8080".into()
```
peterschwarz: What should happen in this case now?

RyanLassigBanks: Nothing, based on our conversation yesterday, it doesn't seem like `Endpoint` should check for the protocol scheme

PR #153, opened 2020-01-09
Update Future and actix libraries

RyanLassigBanks:
PR to update futures library from 0.1 to 0.3. Consequently actix-web has also been updated to 2.0

### DIFF
--- a/daemon/src/rest_api/mod.rs
+++ b/daemon/src/rest_api/mod.rs
@@ -34,8 +34,8 @@ use actix_web::{
     error::{Error as ActixError, ErrorBadRequest, ErrorInternalServerError},
     web, App, FromRequest, HttpRequest, HttpServer, Result,
 };
+use futures::executor::block_on;
 use futures::future;
-use futures::Future;
 use serde::{Deserialize, Serialize};
 
 const SYNC_ARBITER_THREAD_COUNT: usize = 2;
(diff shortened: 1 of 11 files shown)


daemon/src/rest_api/mod.rs
```diff
@@ -106,12 +106,12 @@ impl FromRequest for AcceptServiceIdParam {
...
 pub struct RestApiShutdownHandle {
-    do_shutdown: Box<dyn Fn() -> Result<(), RestApiServerError> + Send>,
+    server: dev::Server,
 }
 
 impl RestApiShutdownHandle {
-    pub fn shutdown(&self) -> Result<(), RestApiServerError> {
-        (*self.do_shutdown)()
+    pub fn shutdown(&self) {
+        block_on(self.server.stop(true));
```
peterschwarz: I'm not sure this should block, since it is called in the ctrl-c handle

RyanLassigBanks: The `do_shutdown` callback that's being replaced by `Server` also blocks by calling `wait`. `wait` doesn't exist in 0.3, so we call `block_on` instead.

peterschwarz: The previous way may have been wrong as well. This should just signal shutdown, and signal other items to shutdown.  We can fix this in a future PR, but should create an issue for it.

PR #156, opened 2020-01-10
Build gridd with experimental features in splinter example

rbuysse:


rbuysse:
Requires #155

PR #157, opened 2020-01-13
Implement Splinter Batch Submitter

peterschwarz:


daemon/src/sawtooth/batch_submitter.rs
```diff
@@ -40,21 +42,33 @@ impl SawtoothBatchSubmitter {
...
+        match $try_expr {
+            Ok(res) => res,
+            Err(err) => return futures::future::err(err).boxed(),
+        }
+    };
+}
+
 impl BatchSubmitter for SawtoothBatchSubmitter {
-    fn submit_batches(&self, msg: SubmitBatches) -> Result<BatchStatusLink, RestApiResponseError> {
+    fn submit_batches(
```
RyanLassigBanks: Since all the futures stuff is updated, this could be `async fn submit_batches(..) -> Result<..>`

peterschwarz: async traits are not yet supported

peterschwarz: async traits are not yet supported

PR #158, opened 2020-01-13
Add service id option to Grid CLI

eloaverona:
Adds service id option for all agents, organization, schema and products CLI commands.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -62,6 +62,8 @@ fn run() -> Result<(), CliError> {
         (@arg wait: --wait +takes_value "How long to wait for transaction to be committed")
         (@arg key: -k +takes_value "base name for private key file")
         (@arg verbose: -v +multiple "Log verbosely")
+        (@arg service_id: --service_id +takes_value "The ID of the service the payload should be \
+            sent to; required if running on Splinter. Format <circuit-id>::<service-id>")
         (@subcommand agent =>
             (about: "Update or create agent")
             (@setting SubcommandRequiredElseHelp)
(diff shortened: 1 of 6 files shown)


cli/src/main.rs
```diff
@@ -62,6 +62,8 @@ fn run() -> Result<(), CliError> {
         (@arg wait: --wait +takes_value "How long to wait for transaction to be committed")
         (@arg key: -k +takes_value "base name for private key file")
         (@arg verbose: -v +multiple "Log verbosely")
+        (@arg service_id: --service_id +takes_value "If running on splinter, provide the ID for \
+            the service the payload should be sent to")
```
ltseeley: The help message should probably be formatted like: "The ID of the service the payload should be sent to; required if running on splinter."

cli/src/main.rs
```diff
@@ -62,6 +62,8 @@ fn run() -> Result<(), CliError> {
         (@arg wait: --wait +takes_value "How long to wait for transaction to be committed")
         (@arg key: -k +takes_value "base name for private key file")
         (@arg verbose: -v +multiple "Log verbosely")
+        (@arg service_id: --service_id +takes_value "If running on splinter, provide the ID for \
```
chenette: Splinter (per capitalization guidelines).

eloaverona: updated

cli/src/main.rs
eloaverona: updated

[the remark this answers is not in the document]
eloaverona: Would this be confusing? I thought circuit-id is part of the service-id. `<circuit-id>::<service-id>` seems to imply they are separate things the user has to put together.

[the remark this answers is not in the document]
eloaverona: Ok, looking at Andi's example at https://github.com/Cargill/splinter/pull/414 it seems they are separate things the user has to put together. Given that, to call this argument `service-id` might be confusing?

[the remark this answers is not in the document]
eloaverona: I added the "<circuit-id>::<service-id>" format example for now

PR #164, opened 2020-01-15
Replace simple logger with flexi logger

shannynalayna:
This allows for further configuration with the logging messages. This pr filters out tokio and hyper logging messages.

### DIFF
--- a/contracts/pike/src/main.rs
+++ b/contracts/pike/src/main.rs
@@ -64,7 +64,6 @@ fn main() {
 
     let mut log_spec_builder = LogSpecBuilder::new();
     log_spec_builder.default(log_level);
-
     match Logger::with(log_spec_builder.build()).start() {
         Ok(_) => (),
         Err(_) => process::exit(1),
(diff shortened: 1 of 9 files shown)


contracts/pike/src/main.rs
```diff
@@ -53,13 +56,19 @@ fn main() {
...
+        0 => log::LevelFilter::Warn,
+        1 => log::LevelFilter::Info,
+        2 => log::LevelFilter::Debug,
+        _ => log::LevelFilter::Trace,
     };
 
-    logger.expect("Failed to create logger");
+    let mut log_spec_builder = LogSpecBuilder::new();
+    log_spec_builder.default(log_level);
+    log_spec_builder.module("hyper", log::LevelFilter::Warn);
```
eloaverona: hyper an tokio are not dependencies for pike nor the CLI, is it necessary to add the log filter here?

shannynalayna: I figured it would be better to cast a wide net, and this keeps it consistent with the other places the logger was added.

peterschwarz: It does create some confusion as to whether or not hyper, etc is involved at this location.

shannynalayna: Addressed by taking these filters out of the cli and pike contract.

PR #166, opened 2020-01-15
Add newline to key files generated by grid CLI

ltseeley:
This makes the output nicer when calling `cat` on a key file.

Also updates the reading of key files to trim any whitespace, which is
necessary now that key files contain a newline.

### DIFF
--- a/cli/src/actions/mod.rs
+++ b/cli/src/actions/mod.rs
@@ -23,3 +23,14 @@ pub mod keygen;
 pub mod organizations;
 pub mod products;
 pub mod schemas;
+
+use std::fs::File;
+use std::io::Write;
+
+use crate::error::CliError;
+
+/// Write the given hex string to the given file, appending a newline at the end.
+fn write_hex_to_file(hex: &str, file: &mut File) -> Result<(), CliError> {
+    writeln!(file, "{}", hex)?;
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


cli/src/actions/mod.rs
```diff
@@ -23,3 +23,14 @@ pub mod keygen;
...
 pub mod schemas;
+
+use std::fs::File;
+use std::io::Write;
+
+use crate::error::CliError;
+
+/// Write the given hex string to the given file, appending a newline at the end.
+fn write_hex_to_file(hex: &str, file: &mut File) -> Result<(), CliError> {
+    file.write(format!("{}{}", hex, "\n").as_bytes())?;
```
peterschwarz: you could use `writeln!(file, hex)` instead.

ltseeley: Updated

cli/src/actions/mod.rs
```diff
@@ -23,3 +23,14 @@ pub mod keygen;
...
 pub mod products;
 pub mod schemas;
+
+use std::fs::File;
+use std::io::Write;
+
+use crate::error::CliError;
+
+/// Write the given hex string to the given file, appending a newline at the end.
+fn write_hex_to_file(hex: &str, file: &mut File) -> Result<(), CliError> {
```
vaporos: My preference would be to avoid wrapper functions like this that don't really do anything.

PR #167, opened 2020-01-16
CLI cleanup

eloaverona:
This PR:
- Changes the `active` argument in the CLI command `agent create` and `agent update` to be flags: `--active` `--not-active`. Orginially the argument took a boolean value
- Changes the `--roles` argument  in the CLI command `agent create` and `agent update` so it accepts comma separated values
- Changes the `verbose` flag to be global, meaning it can be used in any of the subcommands.
- Adds a `quiet` global flag to omit the output
- Changes some printlns to debug level log

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -73,7 +73,7 @@ fn run() -> Result<(), CliError> {
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
                 (@arg active: "Is user active")
-                (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
+                (@arg role: --role +takes_value +use_delimiter +multiple "Roles assigned to agent")
                 (@arg metadata: --metadata +takes_value +multiple
                     "Comma-separated key value pairs stored in metadata")
             )
@@ -82,7 +82,7 @@ fn run() -> Result<(), CliError> {
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
                 (@arg active: "Is user active")
-                (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
+                (@arg role: --role +takes_value +use_delimiter +multiple "Roles assigned to agent")
                 (@arg metadata: --metadata +takes_value +multiple
                     "Comma-separated key value pairs stored in metadata")
             )
@@ -252,7 +252,7 @@ fn run() -> Result<(), CliError> {
                     .with_public_key(m.value_of("public_key").unwrap().into())
                     .with_active(m.is_present("active"))
                     .with_roles(
-                        m.values_of("roles")
+                        m.values_of("role")
                             .unwrap_or_default()
                             .map(String::from)
                             .collect::<Vec<String>>(),
@@ -269,7 +269,7 @@ fn run() -> Result<(), CliError> {
                     .with_public_key(m.value_of("public_key").unwrap().into())
                     .with_active(m.is_present("active"))
                     .with_roles(
-                        m.values_of("roles")
+                        m.values_of("role")
                             .unwrap_or_default()
                             .map(String::from)
                             .collect::<Vec<String>>(),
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -71,17 +72,23 @@ fn run() -> Result<(), CliError> {
...
                     "Comma-separated key value pairs stored in metadata")
             )
             (@subcommand update =>
                 (about: "Update an agent")
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
-                (@arg active: "Is user active")
-                (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
+                (@arg active: --active conflicts_with[not_active] required_unless[not_active]
+                    "Set user as active")
```
dplumb94: user -> agent

cli/src/main.rs
```diff
@@ -71,17 +72,23 @@ fn run() -> Result<(), CliError> {
...
                 (@arg metadata: --metadata +takes_value +multiple
                     "Comma-separated key value pairs stored in metadata")
             )
             (@subcommand update =>
                 (about: "Update an agent")
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
-                (@arg active: "Is user active")
-                (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
+                (@arg active: --active conflicts_with[not_active] required_unless[not_active]
```
dplumb94: should this default to active? I think you wouldn't normally create an inactive agent, right?

dplumb94: Or you could just have a `--inactive` flag, and when it is not used the agent created with active=true

cli/src/main.rs
eloaverona: updated

cli/src/main.rs
dplumb94: ^ this comment should be on the `create` subcommand

eloaverona: I talked to @agunde406 about that, she said it would be better if there were no default value for this.

cli/src/main.rs
```diff
@@ -72,17 +72,23 @@ fn run() -> Result<(), CliError> {
                 (about: "Create an agent")
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
-                (@arg active: "Is user active")
                 (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
+                (@arg active: --active conflicts_with[not_active] required_unless[not_active]
+                    "Set agent as active")
+                (@arg not_active: --not_active conflicts_with[active] required_unless[active]
```
vaporos: Should we call this --inactive?

cli/src/main.rs
vaporos: Agree with no default

cli/src/main.rs
```diff
@@ -72,23 +72,23 @@ fn run() -> Result<(), CliError> {
...
             (@subcommand update =>
                 (about: "Update an agent")
                 (@arg org_id: +takes_value +required "organization ID")
                 (@arg public_key: +takes_value +required "public key")
-                (@arg roles: --roles +takes_value +multiple "Roles assigned to agent")
                 (@arg active: --active conflicts_with[not_active] required_unless[not_active]
                     "Set agent as active")
                 (@arg not_active: --not_active conflicts_with[active] required_unless[active]
                     "Set agent as not active")
+                (@arg roles: --roles +takes_value +use_delimiter +multiple "Roles assigned to agent")
```
vaporos: This should be called --role, with the expectation that multiple arguments are provided (--role a --role b), one per role. I think it's ok though if it also takes a comma separated list in additon.

cli/src/main.rs
eloaverona: updated

cli/src/main.rs
eloaverona: updated to be named --role. And the command accepts both, comma separated values or multiple arguments

PR #174, opened 2020-01-24
Add examples/splinter/README.md

ltseeley:

### DIFF
--- a/examples/splinter/README.md
+++ b/examples/splinter/README.md
@@ -1,0 +1,334 @@
+# Running Hyperledger Grid on Splinter
+
+This document shows how to set up a Grid-on-Splinter environment that runs in a
+set of Docker containers.
+
+The example Splinter docker-compose file creates a network with three nodes
+(alpha, beta, and gamma) that can be used for demos or application development.
+This environment includes the Pike, Product, and Schema smart contracts.
+
+- **Pike** handles organization and identity permissions with Sabre, a smart
+  contract engine that is included in the Splinter scabbard service.
+- **Product** provides a way to share GS1-compatible product data (items that
+  are transacted, traded, or referenced in a supply chain).
+- **Schema** provides a reusable, standard approach to defining, storing, and
+  consuming the product properties. Property definitions are collected into a
+  Schema data type that defines all the possible properties for an item.
+
+
+## Prerequisites
+
+- Docker Engine
+- Docker Compose
+
+
+## Important Note
+
+The example `docker-compose.yaml` file uses experimental Splinter features that
+have not been thoroughly tested or documented.
+
+
+## Set Up and Run Grid
+
+1. Clone the [Hyperledger Grid repository](https://github.com/hyperledger/grid)
+   ([https://github.com/hyperledger/grid](https://github.com/hyperledger/grid)).
+2. Navigate to the grid root directory and start the Grid Docker containers.
+
+   `$ docker-compose -f examples/splinter/docker-compose.yaml up --build`
+
+   This docker-compose file creates a network with two nodes (alpha and beta)
+   that includes the Pike, Schema, and Product smart contracts.
+
+
+## Create a Circuit
+
+To create a circuit, a user on one node proposes a new circuit that includes one
+or more other nodes. When the other nodes accept the circuit proposal, the
+circuit is created.
+
+1. Get the gridd public key from the `gridd-alpha` container. You will need this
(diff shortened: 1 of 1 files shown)


examples/splinter/README.md
```diff
@@ -0,0 +1,337 @@
...
+have not been thoroughly tested or documented.
+
+
+## Set Up and Run Grid
+
+1. Clone the [Hyperledger Grid repository](https://github.com/hyperledger/grid)
+   ([https://github.com/hyperledger/grid](https://github.com/hyperledger/grid)).
+2. Navigate to the grid root directory and start the Grid Docker containers.
+
+   `$ docker-compose -f examples/splinter/docker-compose.yaml up --build`
```
chenette: Example commands with a single line should be formatted as code blocks (with three backticks) instead of in-line code. For example:

> \`\`\`
> $ docker-compose -f examples/splinter/docker-compose.yaml up --build
> \`\`\`

examples/splinter/README.md
```diff
@@ -0,0 +1,337 @@
...
+         - ["peer_services", "[\"grid-scabbard-a\"]"]
+   members:
+     - node_id: alpha-node-000
+       endpoint: tls://splinterd-alpha:8044
+     - node_id: beta-node-000
+       endpoint: tls://splinterd-beta:8044
+   authorization_type: Trust
+   durability: NoDurability
+   circuit_management_type: grid
+   ```
```
chenette: Unfortunately, Markdown doesn't allow "nested" highlighting or formatting in a code block, so  it's hard to see exactly what content to replace (and where). Also, because it takes five copy-paste operations to provide the right content, errors are extremely likely.

My recommended change: First, set an environment variable for the grid-alpha public key, use `cat` to create the  "starter" YAML content in a temporary file (simply copy-paste the entire content), then use `sed` to replace the placeholder key with the actual key and store the output in `circuit.yaml`.  For example:

> `root@splinterd-alpha:/#  GRID-ALPHA-KEY='< actual-grid-alpha-public-key >'`
> 
> `root@splinterd-alpha:/#  cat > circuit.in`
> `circuit_id: my-grid-circuit`
>  `roster:`
>     ` . `
>     ` . `
>     ` . `
> 
> `root@splinterd-alpha:/#  sed -e s/alpha-grid public key/$GRID-ALPHA-KEY/ < circuit.in > circuit.yaml`
>

Feel free to simplify, if possible.

examples/splinter/README.md
```diff
@@ -0,0 +1,337 @@
...
+   in the `~/.grid/keys/` directory.
+
+3. Get the alpha-agent public key. You will need this key when creating a new
+   Grid product in step 5.
+
+   `root@gridd-alpha:/# cat ~/.grid/keys/alpha-agent.pub`
+
+4. Create a new organization, `myorg`.
+
+   `root@gridd-alpha:/# grid -k alpha-agent --url 'http://localhost:8080' --service-id 'my-grid-circuit::grid-scabbard-a' organization create 314156 myorg '123 main street' --metadata gs1_company_prefixes=314156`
```
chenette: The final 6 in the organization ID/prefix should be a 9 (here and on the other two lines).  😄

examples/splinter/README.md
```diff
@@ -0,0 +1,337 @@
...
+   agent with the “admin” role for the new organization (this agent’s public key
+   is derived from the private key used to sign the transaction.) The service ID
+   includes the circuit name and the scabbard service name for the alpha node.
+
+5. Update the agent's permissions (Pike roles) to allow creating, updating, and
+   deleting Grid products.
+
+   **Note**: Replace "<alpha-agent public_key>" with the agent's public key from
+   step 3.
+
```
chenette: This note is wrong (not needed) now that the example uses `cat` to include the contents of the key file.  

P.S. If you're wondering why Markdown ate the contents between the double quotes: A pair of in-line < angle brackets > is interpreted an HTML directive unless the brackets are escaped.

examples/splinter/README.md
peterschwarz: No

examples/splinter/README.md
peterschwarz: No

agunde406: This can't change because product needs this specific values to pass gs1 checks

examples/splinter/README.md
ltseeley: Removed

PR #176, opened 2020-01-27
Minor Grid-on-Splinter doc fixes

peterschwarz:


examples/splinter/README.md
```diff
@@ -282,7 +277,7 @@ circuit is created.
    --role admin
    ```
 
-6. Use `cat` to create a product definition file, `prodcut.yaml`, using the
+5. Use `cat` to create a product definition file, `prodcut.yaml`, using the
```
chenette: Would you be so kind as to correct the file-name typo, as long as you're changing this line?  Thanks!

peterschwarz: Added a commit for this.

PR #178, opened 2020-01-27
 Log as debug when unknown state change received 

ltseeley:
When a state change is received for a key that gridd does not recognize,
this should be logged as a debug message rather than an error. It's
perfectly normal for state to change in ways that aren't directly
relevant to the grid daemon.

### DIFF
--- a/daemon/src/event/db_handler.rs
+++ b/daemon/src/event/db_handler.rs
@@ -45,8 +45,9 @@ use crate::database::{
 };
 
 use super::{
-    CommitEvent, EventError, EventHandler, StateChange, GRID_PRODUCT, GRID_SCHEMA, PIKE_AGENT,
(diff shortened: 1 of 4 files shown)


daemon/src/event/db_handler.rs
```diff
@@ -439,22 +441,22 @@ fn state_change_to_db_operation(
...
 
-                Ok(DbInsertOperation::Products(
+                Ok(Some(DbInsertOperation::Products(
                     product_tuple.0,
                     product_tuple.1,
-                ))
+                )))
+            }
+            _ => {
+                debug!("received state change for unknown key: {}", key);
```
vaporos: Suggest "unknown address" instead of "unknown key".

I'm not sure this should even be logged, because we often enable debug by default and this could get quite verbose. Maybe it should filter out splinter prefixes with no logging.

ltseeley: Fixed (both)

PR #182, opened 2020-01-30
Add section on circuit scope to Grid on Splinter README

agunde406:
Signed-off-by: Andrea Gunderson <agunde@bitwise.io>

### DIFF
--- a/examples/splinter/README.md
+++ b/examples/splinter/README.md
@@ -327,6 +327,40 @@ circuit is created.
    product list
    ```
 
+## Demonstrate Circuit Scope
+
+If a node is not a part of a circuit, that node cannot access information about
+that circuit or any transactions that occur on that circuit.
+
+Use the following steps to demonstrate that the third node in the network
+(gamma-node-000) cannot see the circuit between alpha and beta, even when it
+participates in a new multi-party circuit with those nodes.
+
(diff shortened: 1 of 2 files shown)


integration/docker-compose.yaml
```diff
@@ -198,8 +198,6 @@ services:
       INTEGRATION_TEST_URL: http://gridd:8080
     expose:
       - 8080
-    ports:
```
RyanLassigBanks: Could you elaborate on why this needs to be removed?

agunde406: updated

examples/splinter/README.md
```diff
@@ -327,6 +327,39 @@ circuit is created.
...
+   root@splinterd-gamma:/# splinter circuit list --url http://splinterd-gamma:8085
+   CIRCUIT ID                                                 | CIRCUIT MANAGEMENT TYPE
+   -----------------------------------------------------------------------------------
+   ```
+
+Final note: Splinter strictly enforces privacy for all information on a
+circuit, including participants, available smart contracts, and transactions
+performed by the participants using those smart contracts.
+
+For example, if gamma creates a circuit with alpha and beta, then uploads the
```
peterschwarz: For this scenario to make sense, gamma needs to create separate two-party circuits with alpha and beta. This isn't clear from the wording here.

PR #188, opened 2020-02-05
Add env variables for key and url CLI args

shannynalayna:
Adds environment variables to the Splinter example docker-compose file for the key and url CLI arguments, as well as checks within the Grid CLI for these values. 

Also updates the README to remove the use of -k and --url in the commands for gridd.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -31,6 +31,8 @@ mod splinter;
 mod transaction;
 mod yaml_parser;
 
+use std::env;
+
 use clap::ArgMatches;
 use flexi_logger::{LogSpecBuilder, Logger};
 use grid_sdk::protocol::pike::{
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -51,6 +53,9 @@ use actions::admin;
 const APP_NAME: &str = env!("CARGO_PKG_NAME");
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
+const KEY_FILE_BASE: &str = "KEY_FILE_BASE";
```
vaporos: This and endpoint should be prefixed with GRID_. Maybe GRID_DAEMON_KEY and GRID_DAEMON_ENDPOINT.

shannynalayna: Addressed.

PR #189, opened 2020-02-06
Add exponential back-off to status check

RyanLassigBanks:
`get_node_id` now tries indefinitely to check the node status doubling
it's wait time after each failed attempt with a maximum wait time of 300
seconds.

### DIFF
--- a/daemon/src/splinter/app_auth_handler/node.rs
+++ b/daemon/src/splinter/app_auth_handler/node.rs
@@ -17,6 +17,8 @@
 use serde_json::Value;
 use std::error::Error;
 use std::fmt;
+use std::thread;
+use std::time::Duration;
 
 #[derive(Debug, PartialEq)]
 pub struct GetNodeError(pub String);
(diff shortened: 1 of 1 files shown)


daemon/src/splinter/app_auth_handler/node.rs
```diff
@@ -51,3 +50,24 @@ pub fn get_node_id(splinterd_url: String) -> Result<String, GetNodeError> {
...
+    let mut wait_time = 1;
+    loop {
+        match reqwest::blocking::get(uri) {
+            Ok(res) => {
+                return res.json().map_err(|err| {
+                    GetNodeError(format!("Failed to parse response body: {}", err))
+                });
+            }
+            Err(err) => {
+                error!("Failed to get set up request: {}", err);
```
agunde406: Since these are not actually raising an error, maybe these should be warn messages

RyanLassigBanks: changed to warnings

daemon/src/splinter/app_auth_handler/node.rs
```diff
@@ -51,3 +50,28 @@ pub fn get_node_id(splinterd_url: String) -> Result<String, GetNodeError> {
...
+    let mut wait_time = 1;
+    loop {
+        match reqwest::blocking::get(uri) {
+            Ok(res) => {
+                return res.json().map_err(|err| {
+                    GetNodeError(format!("Failed to parse response body: {}", err))
+                });
+            }
+            Err(err) => {
+                warn!("Failed to get set up request: {}", err);
```
peterschwarz: This is a strange message.  I would suggest changing it to "Unable to get splinter status"

PR #190, opened 2020-02-06
Display output for keygen commands

eloaverona:
This PR fix some of the log levels in the Grid CLI. Now the `grid keygen` and `grid admin keygen` CLI commands to display an output unless the `--quiet` flag is present.


cli/src/main.rs
```diff
@@ -217,7 +217,9 @@ fn run() -> Result<(), CliError> {
 
     let matches = app.get_matches();
 
-    let log_level = if matches.is_present("quiet") {
```
peterschwarz: Should this only be set based on the verbose level now?

PR #191, opened 2020-02-07
Service-id environment variable

shannynalayna:
Adds a `GRID_SPLINTER_SERVICE_ID` environment variable to the gridd alpha and beta docker containers for the Splinter example. Also changes around the actions functions to expect Option<String> instead of Option<&str> because the environment variable gets returned as Option<String> but is unable to directly be converted using `deref` because the value being created is temporary when fetched using `env::var`.

### DIFF
--- a/examples/splinter/docker-compose.yaml
+++ b/examples/splinter/docker-compose.yaml
@@ -136,6 +136,7 @@ services:
     environment:
       GRID_DAEMON_KEY: "alpha-agent"
       GRID_DAEMON_ENDPOINT: "http://gridd-alpha:8080"
+      GRID_SERVICE_ID: "my-grid-circuit::grid-scabbard-a"
     entrypoint: |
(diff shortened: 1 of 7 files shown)


examples/splinter/docker-compose.yaml
```diff
@@ -136,6 +136,7 @@ services:
     environment:
       GRID_DAEMON_KEY: "alpha-agent"
       GRID_DAEMON_ENDPOINT: "http://gridd-alpha:8080"
+      GRID_SERVICE_ID: "my-grid-circuit::grid-scabbard-a"
```
agunde406: Can we make this environment variable splinter specific somehow?

shannynalayna: Addressed! Changed to `GRID_SPLINTER_SERVICE_ID`

PR #192, opened 2020-02-07
Change CommitEvent service_id to Option<String>

yarrkov:
Using docker-compose.yaml and grid-cli to create a test organization, I ended up with an empty string as service_id in the PostgreSQL db. gridd expects service_id to be null in places like <https://github.com/hyperledger/grid/blob/master/daemon/src/database/helpers/organizations.rs#L72>. Thus the gridd REST API returns nothing. This seems to fix the issue.


yarrkov:
I just screwed up with git here. I'll see if I can manage to fix that.

PR #200, opened 2020-02-20
Remove last mentions of my-grid-circuit

agunde406:
The circuit create command no longer lets you chose the name
of your circuit so this commit removes or replaces all mentions
of my-grid-circuit with an env variable CIRCUIT_ID.

Also updates the version of XO that is deployed.

### DIFF
--- a/examples/splinter/README.md
+++ b/examples/splinter/README.md
@@ -90,18 +90,27 @@ circuit is created.
       --service-peer-group grid-scabbard-a,grid-scabbard-b
    ```
 
-5. Check the results by displaying the list of proposals. Then use the circuit
-   ID to view the details of the new proposal.
+5. Check the results by displaying the list of proposals. The following example
+   sets the CIRCUIT_ID environment variable; this environment variable is for
+   the purposes of this procedure and is not used directly by the `splinter`
+   CLI commands.
+
+   Set CIRCUIT_ID based on the output of the `proposals` subcommand; for
+   example:
 
    ```
    root@splinterd-alpha:/# splinter circuit proposals --url http://splinterd-alpha:8085
-   ID              MANAGEMENT MEMBERS
-   <circuit-id>    grid       alpha-node-000;beta-node-000
+   ID                                      MANAGEMENT MEMBERS
+   01234567-0123-0123-0123-012345678901    grid       alpha-node-000;beta-node-000
    ```
 
    ```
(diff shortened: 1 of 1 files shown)


examples/splinter/README.md
```diff
@@ -309,7 +309,7 @@ circuits.
    --owner $(cat /root/.splinter/keys/gridd.pub) \
    -k gridd \
    -U 'http://splinterd-alpha:8085' \
-   --service-id my-grid-circuit::grid-scabbard-a
+   --service-id <circuit-id> ::grid-scabbard-a
```
ltseeley: Looks like you may have introduced a space before `::`

examples/splinter/README.md
```diff
@@ -90,18 +90,26 @@ circuit is created.
       --service-peer-group grid-scabbard-a,grid-scabbard-b
    ```
 
-5. Check the results by displaying the list of proposals. Then use the circuit
-   ID to view the details of the new proposal.
+5. Check the results by displaying the list of proposals. For the following
+   commands, we set the CIRCUIT_ID environment variable; this environment
+   variable is for the purposes of this procedure and is not used directly by
+   the splinter CLI commands.
+
```
chenette: If you're referring to the splinter command aka CLI, it should be in fixed-width font. (The product is Splinter with a capital S.)

Also, the doc style guide recommends avoiding "we". Instead, I suggest: "The following example sets the ..."

examples/splinter/README.md
```diff
@@ -90,18 +90,26 @@ circuit is created.
...
+   Set CIRCUIT_ID based on the output of the proposals subcommand; for example:
```
chenette: `proposals` should be be in fixed-width font.

examples/splinter/README.md
```diff
@@ -202,13 +216,14 @@ circuit is created.
    This command generates two files, `alpha-agent.priv` and `alpha-agent.pub`,
    in the `~/.grid/keys/` directory.
 
-3. Set an environment variable with the service_id. The commands below will
-check this variable to determine which circuit and service the command should
-be run against. An alternative to using the environment variable is to pass the
-service_id via the `--service-id` argument in each of these commands.
+3. Set an environment variable with the service_id. Use the circuit id of the
```
chenette: Change here and elsewhere: 
- service ID (not service_id, because the option has a dash, not an underscore
- circuit ID

PR #209, opened 2020-03-11
Document products endpoints

shannynalayna:
Adds the necessary information for the products endpoints,
including /products and /products/{product_id}, as well as
the product definition to the openapi spec.

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -1,5 +1,6 @@
 # Copyright 2017 Intel Corporation
 # Copyright 2019 Bitwise IO, Inc.
+# Copyright 2020 Cargill Incorporated
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


daemon/openapi.yaml
```diff
@@ -422,6 +422,60 @@ paths:
...
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/Product"
+        "400":
+          $ref: "#/components/responses/400BadRequest"
+        "500":
+          $ref: "#/components/responses/500ServerError"
+        "503":
+          $ref: "#/components/responses/503ServiceUnavailable"
+  "/product/{product_id}":
```
peterschwarz: Does this route have a 404 Response?

shannynalayna: Yes, it does. Added.

daemon/openapi.yaml
```diff
@@ -422,6 +422,62 @@ paths:
             $ref: "#/components/responses/500ServerError"
           "503":
             $ref: "#/components/responses/503ServiceUnavailable"
+  /product:
```
dplumb94: It looks like handler for list products can take an optional service ID query parameter? If that is true it should be added.

shannynalayna: Totally right, added!

shannynalayna:
@dplumb94 I added a component schema entry for the `properties` field in the product endpoint response

PR #212, opened 2020-03-13
Add service_id to rest api response objects

shannynalayna:
Adds an optional service id to all the rest api response objects. Also adds a check for this in the existing tests and adds documentation to the api spec.

### DIFF
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -735,6 +735,15 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Metadata"
+        service_id:
+          type: string
+          example: 01234567-0123-0123-0123-012345678901::grid-scabbard-a
+      required:
+        - public_key
+        - org_id
+        - active
+        - roles
+        - metadata
     Organization:
       type: object
       properties:
--- a/daemon/openapi.yaml
+++ b/daemon/openapi.yaml
@@ -972,6 +972,18 @@ components:
               type: string
           lat_long_value:
             $ref: '#/components/schemas/LatLong'
+        service_id:
+          type: string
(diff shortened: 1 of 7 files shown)


daemon/openapi.yaml
```diff
@@ -661,6 +661,9 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Metadata"
+        service_id:
```
peterschwarz: Is there a way to highlight that this is optional?

daemon/src/rest_api/routes/agents.rs
```diff
@@ -29,6 +29,7 @@ pub struct AgentSlice {
     pub active: bool,
     pub roles: Vec<String>,
     pub metadata: JsonValue,
+    pub service_id: Option<String>,
```
peterschwarz: Should these skip serializing if none, and default to none on deserialize if not set? Same with other occurrences.

daemon/src/rest_api/routes/products.rs
```diff
@@ -36,6 +36,7 @@ pub struct ProductSlice {
     pub product_namespace: String,
     pub owner: String,
     pub properties: Vec<ProductPropertyValueSlice>,
+    pub service_id: Option<String>,
```
peterschwarz: Does not look like you added it to the open api spec in this commit.

shannynalayna: The products endpoint isn't currently documented in the openapi spec. It will be, however, once https://github.com/hyperledger/grid/pull/209 is merged.

daemon/openapi.yaml
shannynalayna: I added a `required` field to each component schema that I also had added `service_id` to. This was the best way to represent the optional nature of the service_id

PR #214, opened 2020-03-17
Update circuit id and service ids for new format

agunde406:
The new circuit ID format is a string composed of two, 5
character base62 strings joined by a dash.
Example: aB3dE-7gH0j

The new service ID format is a 4 character base62 string.

Signed-off-by: Andrea Gunderson <agunde@bitwise.io>

### DIFF
--- a/examples/splinter/README.md
+++ b/examples/splinter/README.md
@@ -82,12 +82,12 @@ circuit is created.
       --url http://splinterd-alpha:8085  \
       --node alpha-node-000::tls://splinterd-alpha:8044 \
       --node beta-node-000::tls://splinterd-beta:8044 \
-      --service grid-scabbard-a::alpha-node-000 \
-      --service grid-scabbard-b::beta-node-000 \
+      --service gsAA::alpha-node-000 \
+      --service gsBB::beta-node-000 \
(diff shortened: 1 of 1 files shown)


examples/splinter/README.md
```diff
@@ -82,12 +82,12 @@ circuit is created.
...
       --node beta-node-000::tls://splinterd-beta:8044 \
-      --service grid-scabbard-a::alpha-node-000 \
-      --service grid-scabbard-b::beta-node-000 \
+      --service gsAA::alpha-node-000 \
+      --service gsBB::beta-node-000 \
       --service-type *::scabbard \
       --management grid \
       --service-arg *::admin_keys=$(cat gridd.pub) \
-      --service-peer-group grid-scabbard-a,grid-scabbard-b
+      --service-peer-group gsAA,gsBB
```
desktophero: I ended up using `--service-arg *::peer_services=alpg,brag` prior to seeing this.

They both appear to do the same thing. Is there a recommended approach?

desktophero: oh - never mind. I see what happens when I group. My service peers show as that literal value.

PR #215, opened 2020-03-19
Add view for the AddProductForm

dnewh:
This adds the view for the form to add products in the Product sapling for a selected group of services. It is not currently hooked up to gridd. This PR also includes several components and config updates required for building this form.

## Testing

* Run `export 'CARGO_ARGS=-- --features experimental'` from the Grid directory
* Run `docker-compose -f grid-ui/docker/docker-compose.yaml build gridd grid-db splinter-db splinterd`
* Run `docker-compose -f grid-ui/docker/docker-compose.yaml up gridd grid-db splinter-db splinterd`
* Navigate to the `grid/grid-ui` directory
* Run `npm install`
* Navigate to any of the sapling directories that are required for testing, in this case `grid/grid-ui/saplings/product`, and run `npm install && npm run deploy` in them.
* Navigate back to the `grid/grid-ui` directory and run `npm run start`
* Once the app opens in the browser navigate to the profile sapling by clicking the icon in the bottom of the nav bar

### DIFF
--- a/grid-ui/.prettierignore
+++ b/grid-ui/.prettierignore
@@ -14,3 +14,4 @@
 
 saplings/*
 **/build/*
+sapling-dev-server/*
--- a/grid-ui/saplings/product/src/App.js
+++ b/grid-ui/saplings/product/src/App.js
@@ -36,7 +36,7 @@ library.add(faCaretUp, faCaretDown, faCheck, faPenSquare, faChevronLeft);
 function App() {
   return (
     <ServiceProvider>
-      <div className="product-app">
+      <div id="product-sapling" className="product-app">
         <FilterBar />
         <Router>
           <Switch>
--- a/grid-ui/saplings/product/src/App.js
+++ b/grid-ui/saplings/product/src/App.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import { library } from '@fortawesome/fontawesome-svg-core';
 import {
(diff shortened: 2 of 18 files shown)


grid-ui/.prettierignore
```diff
@@ -14,3 +14,4 @@
 
 saplings/*
 build/*
+sapling-dev-server/product
```
dplumb94: This should probably just be `sapling-dev-server/*` since it would be the same case for every sapling.

grid-ui/saplings/product/src/components/MultiStepForm.js
```diff
@@ -0,0 +1,202 @@
+/**
```
dplumb94: Is this the same component from the Profile sapling? If so I think we should put it somewhere that it can be reused?

grid-ui/saplings/product/src/App.js
```diff
@@ -14,28 +14,50 @@
...
+import React, { useState } from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { library } from '@fortawesome/fontawesome-svg-core';
 import {
   faCaretUp,
   faCaretDown,
   faCheck,
-  faPenSquare
+  faPenSquare,
+  faPlus
```
dplumb94: Right now I am using the Library pattern for fa icons here:
https://www.npmjs.com/package/@fortawesome/react-fontawesome#build-a-library-to-reference-icons-throughout-your-app-more-conveniently

I think we should continue to do that throughout the app or make a decision not to use it, since having it half-and-half gets confusing.

grid-ui/saplings/product/src/App.js
```diff
@@ -14,28 +14,50 @@
...
+    }
+    return null;
+  }
+
   return (
     <ServiceProvider>
       <div className="product-app">
         <FilterBar />
         <ProductsTable />
+        <button className="fab add-product" type="button" onClick={addProduct}>
```
dplumb94: Can this be included on the ProductsTable component? My current work includes a routerview to switch between the products view and the product info view for a specific component, so conceptually this button would be on the ProductsTable.

grid-ui/saplings/product/src/components/MultiStepForm.js
dnewh: It is not the same component. It has some functionality in common but also some differences in functionality and significantly different styling.

grid-ui/saplings/product/src/App.js
dnewh: sure. updated

grid-ui/saplings/product/src/App.js
dnewh: updated to use the library pattern

grid-ui/.prettierignore
dnewh: updated

dnewh:
@dplumb94 comments 1-5 have been addressed and updated

PR #232, opened 2020-04-17
Reflect node endpoints change in splinter

ltseeley:
Updates Grid to reflect the change in splinter that makes node endpoints
plural. This includes updating the node registry file and adding a
database migration to update the column.


ltseeley:
Closing in favor of #233

PR #239, opened 2020-04-22
Fix routing issue with product sampling

RyanLassigBanks:
Fixes bug that was causing the product sampling to not render the canopy
navbar when navigating directly to localhost:3033/products


RyanLassigBanks:
This issue was resolved in canopy closing

PR #240, opened 2020-04-24
Fix lint issues

dnewh:
This updates cli and daemon imports to pass lint with the most recent
version of Rust.


dnewh:
Updated commit message

PR #246, opened 2020-05-04
Fix smart contract names

shannynalayna:
Replaces underscores with dashes within smart contract family names to adhere to the recently implemented naming requirements, which prevents the use of underscores in smart contract names.


shannynalayna:
Unnecessary as of https://github.com/hyperledger/transact/pull/62

PR #250, opened 2020-05-11
Pull scabbard client from scabbard crate

ltseeley:
Updates gridd's dependencies to pull the scabbard client from the newly
created scabbard crate; the client is no longer provided by the splinter
crate.


ltseeley:
This is no longer needed because of #251

PR #252, opened 2020-05-21
Update Splinter event reactor shutdown

shannynalayna:
Updates the Splinter event reactor shutdown to the latest method
using the `shutdown_signaler` to signal a shutdown rather than
using the deprecated `shutdown` method.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -204,6 +204,8 @@ fn run_splinter(config: GridConfig, connection_pool: ConnectionPool) -> Result<(
         config.endpoint().clone(),
     )?;
 
+    let reactor_shutdown_signaler = reactor.shutdown_signaler();
+
     let ctrlc_triggered = AtomicBool::new(false);
     ctrlc::set_handler(move || {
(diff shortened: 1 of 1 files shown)


daemon/src/main.rs
```diff
@@ -224,8 +224,17 @@ fn run_splinter(config: GridConfig, connection_pool: ConnectionPool) -> Result<(
         })
         .and_then(|res| res.map_err(DaemonError::from))?;
 
-    if let Err(err) = reactor.shutdown() {
-        error!("unable to shutdown splinter event reactor: {}", err)
+    let reactor_shutdown_signaler = reactor.shutdown_signaler();
+
+    if let Err(err) = reactor_shutdown_signaler.signal_shutdown() {
+        error!(
```
agunde406, line 230: If this errors, will wait_for_shutdown return?

shannynalayna: Updated, so that the signal_shutdown happens in the ctrl-c handler, as it is in the Gameroom daemon

PR #253, opened 2020-05-26
Switch compose files to use Splinter 0.4 images

rbuysse:
Even numbered Splinter branches are considered "stable" so Grid should
use these instead of master.

### DIFF
--- a/examples/splinter/docker-compose.yaml
+++ b/examples/splinter/docker-compose.yaml
@@ -86,7 +86,7 @@ services:
       "
 
   generate-registry:
-    image: splintercommunity/splinter-cli:master
+    image: splintercommunity/splinter-cli:0.4-dev
     volumes:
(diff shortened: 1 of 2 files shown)


examples/splinter/docker-compose.yaml
```diff
@@ -191,14 +191,14 @@ services:
         "
 
   scabbard-cli-alpha:
-    image: splintercommunity/scabbard-cli:experimental
+    image: splintercommunity/scabbard-cli:0.4-experimental
```
peterschwarz, line 194: Is grid using any scabbard CLI features that are not stable? @ltseeley

vaporos: Yes, this is probably true. Lets do that change separately though, as this change is functionally equiv. given branching changes and moving to stable is different.

PR #289, opened 2020-08-19
Refactor DB

dnewh:
This PR takes the first steps toward refactoring the Grid DB towards following the modular, backend-agnostic store pattern established in [Splinter](https://github.com/Cargill/splinter). This moves the migrations and commit store/operations to the Grid SDK and updated the daemon handler and CLI to use these new implementations. Additional modules can be added or converted following a similar pattern. This new pattern provides the benefits of allowing stores to be modularized and allows operations to be DB backend agnostic though currently only the postgres support is implemented.


dnewh:
Big thanks to @RyanLassigBanks for help with debugging this PR!

PR #306, opened 2020-08-25
Location smart contract

RyanLassigBanks:
Adds implementation for location smart contract

### DIFF
--- a/cli/Dockerfile
+++ b/cli/Dockerfile
@@ -15,11 +15,11 @@
 FROM hyperledger/grid-dev:v3 as grid-cli-builder
 
 # Copy over Cargo.toml files
-COPY Cargo.toml /build/Cargo.toml
 COPY cli/Cargo.toml /build/cli/Cargo.toml
 COPY daemon/Cargo.toml /build/daemon/Cargo.toml
 COPY sdk/Cargo.toml /build/sdk/Cargo.toml
 
+COPY contracts/location/Cargo.toml /build/contracts/location/Cargo.toml
 COPY contracts/pike/Cargo.toml /build/contracts/pike/Cargo.toml
 COPY contracts/product/Cargo.toml /build/contracts/product/Cargo.toml
 COPY contracts/schema/Cargo.toml /build/contracts/schema/Cargo.toml
@@ -43,7 +43,7 @@ FROM hyperledger/sawtooth-sabre-cli
 ARG CARGO_ARGS
 RUN echo "CARGO_ARGS = '$CARGO_ARGS'" > CARGO_ARGS
 
-COPY --from=grid-cli-builder /build/target/debian/grid-cli_*.deb /tmp
+COPY --from=grid-cli-builder /build/cli/target/debian/grid-cli_*.deb /tmp
 
 RUN apt-get update \
  && dpkg --unpack /tmp/grid-cli*.deb \
--- a/contracts/location/src/addressing.rs
+++ b/contracts/location/src/addressing.rs
@@ -1,0 +1,48 @@
+// Copyright 2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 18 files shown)


cli/Dockerfile
```diff
@@ -34,16 +34,16 @@ COPY sdk /build/sdk
 WORKDIR /build/cli
 ARG CARGO_ARGS
 ARG REPO_VERSION
-RUN sed -i -e "0,/version.*$/ s/version.*$/version\ =\ \"${REPO_VERSION}\"/" Cargo.toml
-RUN cargo deb --deb-version $REPO_VERSION $CARGO_ARGS
+#RUN sed -i -e "0,/version.*$/ s/version.*$/version\ =\ \"${REPO_VERSION}\"/" Cargo.toml
+RUN cargo deb --deb-version 1.0 $CARGO_ARGS
```
vaporos, line 38: This looks unintentional.

contracts/location/src/addressing.rs
```diff
@@ -0,0 +1,59 @@
...
+use crypto::sha2::Sha512;
+
+pub const GRID_NAMESPACE: &str = "621dee";
+const GRID_SCHEMA_NAMESPACE: &str = "01";
+
+const GS1_LOCATION_NAMESPACE: &str = "01";
+
+const LOCATION_NAMESPACE: &str = "04";
+
+const PIKE_NAMESPACE: &str = "cad11d";
```
vaporos, line 25: The pike stuff in this file looks like maybe it should be moved to a pike module outside of location.

contracts/location/src/addressing.rs
```diff
@@ -0,0 +1,59 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crypto::digest::Digest;
+use crypto::sha2::Sha512;
+
+pub const GRID_NAMESPACE: &str = "621dee";
+const GRID_SCHEMA_NAMESPACE: &str = "01";
+
+const GS1_LOCATION_NAMESPACE: &str = "01";
```
vaporos, line 21: Could lessen the consts here by explanation instead of construction of strings. For example:

```
// GS1 Location namespace is constructed from the following pieces:
//     "621dee" (grid) + "01" (Location) "01" + (GS1 Location) "01"
const GS1_LOCATION_NAMESPACE: &str = "621dee0101";
```

contracts/location/src/handler.rs
```diff
@@ -0,0 +1,1668 @@
...
+    family_name: String,
+    family_versions: Vec<String>,
+    namespaces: Vec<String>,
+}
+
+impl LocationTransactionHandler {
+    pub fn new() -> Self {
+        Self {
+            family_name: "grid_location".to_string(),
+            family_versions: vec!["1.0".to_string()],
```
vaporos, line 77: We should start using single digit versions in our contracts -- "1"

contracts/location/src/handler.rs
```diff
@@ -0,0 +1,1668 @@
...
+        LocationNamespace::GS1 => StateNamespace::GS1,
+    };
+
+    let location = LocationBuilder::new()
+        .with_location_id(payload.location_id().to_string())
+        .with_namespace(namespace)
+        .with_owner(payload.owner().to_string())
+        .with_properties(payload.properties().to_vec())
+        .build()
+        .expect("Failed to build location");
```
vaporos, line 257: There shouldn't be any panics in a smart contract.

contracts/location/src/addressing.rs
RyanLassigBanks: Removed consts that were only being used in the addressing module

contracts/location/src/addressing.rs
RyanLassigBanks: Removed consts that were only being used in the addressing module

dplumb94 (dismissed):

vaporos (approved):

dplumb94 (approved):

PR #307, opened 2020-08-28
Add sqlite dependency

dnewh:
This adds the libsqlite3-dev dependency to the grid-dev image and increments the version number from 2 to 3. This is done in support of refactoring the Grid database functionality.

### DIFF
--- a/ci/grid-dev
+++ b/ci/grid-dev
@@ -24,6 +24,7 @@ RUN apt-get update \
     libpq-dev \
     libssl-dev \
     libsasl2-dev \
+    libsqlite3-dev \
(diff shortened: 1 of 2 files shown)


ci/grid-dev
```diff
@@ -24,6 +24,7 @@ RUN apt-get update \
     libpq-dev \
     libssl-dev \
     libsasl2-dev \
+    libsqlite3-dev \
```
peterschwarz, line 27: Could you add the reason for this addition to the commit message?

dnewh: added

PR #310, opened 2020-09-01
Add organization store implementation

dnewh:
This adds the postgres, sqlite, and in-memory implementations of the organization store following the new pattern. This does not currently integrate the new store into the daemon but that integration will come in a later PR.

### DIFF
--- a/sdk/src/grid_db/organizations/store/diesel/operations/add_organizations.rs
+++ b/sdk/src/grid_db/organizations/store/diesel/operations/add_organizations.rs
@@ -1,0 +1,74 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::OrganizationStoreOperations;
+use crate::grid_db::organizations::store::diesel::{schema::organization, OrganizationStoreError};
+
+use crate::grid_db::commits::MAX_COMMIT_NUM;
+use crate::grid_db::organizations::store::diesel::models::{
+    NewOrganizationModel, OrganizationModel,
+};
+use diesel::{dsl::insert_into, prelude::*, result::Error::NotFound};
+
+pub(in crate::grid_db::organizations::store::diesel) trait OrganizationStoreAddOrganizationsOperation
+{
+    fn add_organizations(
+        &self,
+        orgs: Vec<NewOrganizationModel>,
+    ) -> Result<(), OrganizationStoreError>;
+}
+
+#[cfg(feature = "postgres")]
+impl<'a> OrganizationStoreAddOrganizationsOperation
+    for OrganizationStoreOperations<'a, diesel::pg::PgConnection>
+{
+    fn add_organizations(
+        &self,
+        orgs: Vec<NewOrganizationModel>,
+    ) -> Result<(), OrganizationStoreError> {
+        for org in orgs {
+            let duplicate_org = organization::table
+                .filter(
+                    organization::org_id
+                        .eq(&org.org_id)
+                        .and(organization::service_id.eq(&org.service_id))
+                        .and(organization::end_commit_num.eq(MAX_COMMIT_NUM)),
+                )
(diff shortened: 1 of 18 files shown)


sdk/src/grid_db/organizations/store/error.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use std::error::Error;
+use std::fmt;
+
+#[cfg(feature = "diesel")]
+use crate::database::error;
+
+/// Represents OrganizationStore errors
+#[derive(Debug)]
+pub enum OrganizationStoreError {
```
RyanLassigBanks, line 23: This has a lot of overlap with CommitStoreError. We should consider having a common error module at the root grid_db

RyanLassigBanks (changes requested):

sdk/src/grid_db/organizations/store/error.rs
dnewh: updated

sdk/src/grid_db/organizations/store/diesel/operations/add_organizations.rs
```diff
@@ -0,0 +1,73 @@
...
+use super::OrganizationStoreOperations;
+use crate::grid_db::organizations::store::diesel::{schema::organization, OrganizationStoreError};
+
+use crate::grid_db::commits::MAX_COMMIT_NUM;
+use crate::grid_db::organizations::store::diesel::models::{
+    NewOrganizationModel, OrganizationModel,
+};
+use diesel::{dsl::insert_into, prelude::*, result::Error::NotFound};
+
+pub(in crate::grid_db::organizations) trait OrganizationStoreAddOrganizationsOperation {
```
vaporos, line 24: Seems odd that code at that high level would need this trait. should it be super::super (which is grid_db::oranizations::store::diesel)? (Same comment for other operations.)

sdk/src/store/sqlite.rs
```diff
@@ -34,4 +34,10 @@ impl StoreFactory for SqliteStoreFactory {
     fn get_grid_commit_store(&self) -> Box<dyn crate::grid_db::CommitStore> {
         Box::new(crate::grid_db::DieselCommitStore::new(self.pool.clone()))
     }
+
+    fn get_grid_organization_store(&self) -> Box<dyn crate::grid_db::OrganizationStore> {
```
vaporos, line 38: This feels like its in the wrong commit as its not sqlite-specific.

vaporos (changes requested):

sdk/src/store/sqlite.rs
dnewh: This is part of the implementation of StoreFactory for SqliteStoreFactory

sdk/src/grid_db/organizations/store/diesel/operations/add_organizations.rs
dnewh: updated

vaporos (dismissed):

sdk/src/hex.rs
```diff
@@ -0,0 +1,157 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error::Error;
+use std::fmt::{self, Write};
+
+use serde::de;
+use serde::{Deserializer, Serializer};
+
```
dplumb94, line 20: I think we need full rustdocs for these functions since its part of the Grid SDK

sdk/src/grid_db/organizations/store/memory.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use std::collections::HashMap;
+use std::iter::FromIterator;
+use std::sync::{Arc, Mutex};
+
+use super::OrganizationStore;
+use crate::grid_db::commits::MAX_COMMIT_NUM;
+use crate::grid_db::organizations::store::{error::OrganizationStoreError, Organization};
+
+///Implementation of OrganizationStore that stores Commits in memory. Useful for when
```
dplumb94, line 23: Just a typo in the comment... Commits should be lowercase unless its referring to an actual object

dplumb94 (changes requested):

sdk/src/grid_db/organizations/store/memory.rs
vaporos: There should be a space after /// as well.

sdk/src/hex.rs
dnewh: Added

sdk/src/grid_db/organizations/store/memory.rs
dnewh: Updated

dplumb94 (approved):

vaporos (approved):

RyanLassigBanks (approved):

PR #311, opened 2020-09-03
Product Smart Contract Updates

RyanLassigBanks:
Several updates to the product smart contract to align it with the RFC

### DIFF
--- a/contracts/product/src/handler.rs
+++ b/contracts/product/src/handler.rs
@@ -658,7 +658,7 @@ mod tests {
         fn add_gs1_schema(&self) {
             let properties = vec![
                 PropertyDefinitionBuilder::new()
-                    .with_name("price".into())
+                    .with_name("counter".into())
                     .with_data_type(SchemaDataType::Number)
                     .with_number_exponent(1)
                     .with_required(true)
@@ -1131,8 +1131,8 @@ mod tests {
             .with_string_value("This is a product description".into())
             .build()
             .unwrap();
-        let property_value_price = PropertyValueBuilder::new()
-            .with_name("price".into())
+        let property_value_counter = PropertyValueBuilder::new()
+            .with_name("counter".into())
             .with_data_type(DataType::Number)
             .with_number_value(3)
             .build()
@@ -1140,7 +1140,7 @@ mod tests {
 
         vec![
             property_value_description.clone(),
-            property_value_price.clone(),
+            property_value_counter.clone(),
         ]
     }
 
@@ -1151,8 +1151,8 @@ mod tests {
             .with_string_value("This is a new product description".into())
             .build()
             .unwrap();
-        let property_value_price = PropertyValueBuilder::new()
-            .with_name("price".into())
+        let property_value_counter = PropertyValueBuilder::new()
+            .with_name("counter".into())
             .with_data_type(DataType::Number)
             .with_number_value(4)
             .build()
@@ -1160,7 +1160,7 @@ mod tests {
 
         vec![
             property_value_description.clone(),
-            property_value_price.clone(),
+            property_value_counter.clone(),
         ]
     }
 
--- a/contracts/product/src/handler.rs
+++ b/contracts/product/src/handler.rs
@@ -170,6 +170,46 @@ impl ProductTransactionHandler {
             }
         }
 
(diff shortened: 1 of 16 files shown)


contracts/product/src/handler.rs
```diff
@@ -561,14 +646,51 @@ mod tests {
...
                 (product_1_address, product_list_bytes),
                 (product_2_address, product_list_bytes_copy),
             ])
             .unwrap();
         }
+
+        fn add_gs1_schema(&self) {
+            let properties = vec![
+                PropertyDefinitionBuilder::new()
+                    .with_name("price".into())
```
vaporos, line 661: "price" is a bad example as it isn't something that actually defines a product. Would be good to find a small handful of real examples for this purpose. (Or maybe put a issue on the board for coming back to this.)

RyanLassigBanks: I chose price as field because the mock products already defined a field called price, and deviating would have broke the unit tests. I don't really see how adding accurate data for the mock schema would improve the unit tests, since the product smart contract doesn't/ can't know what an "accurate" product schema looks like. It just knows that there is a schema that it needs to check the products against.

dplumb94 (changes requested):
I got this error running the grid on splinter instructions when submitting a product:

```
splinterd-alpha             | thread 'StaticExecutionAdapter' panicked at 'byte index 70 is out of bounds of `621dee020100000000000000000000000000000000000000000000072338288508800`', /rustc/5c1f21c3b82297671ad3ae1e8c942d2ca92e84f2/src/libcore/str/mod.rs:1920:47
splinterd-alpha             | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
splinterd-alpha             | [2020-09-11 15:23:02.789] T["StaticExecutionAdapter"] INFO [sawtooth_sabre::wasm_executor::wasm_externals] create_collection: 2063
splinterd-alpha             | [2020-09-11 15:23:02.789] T["StaticExecutionAdapter"] INFO [sawtooth_sabre::wasm_executor::wasm_externals] Attempting to get state, addresses: ["621dee020100000000000000000000000000000000000000000000072338288508800"]
```

Also I think the grid on splinter readme needs to be updated to include adding the schema before the product is added

examples/splinter/README.md
```diff
@@ -269,15 +269,38 @@ steps 3 through 10.
...
+   --role can_create_schema \
    --role admin
    ```
 
-6. Use `cat` to create a product definition file, `product.yaml`, using the
+6. Use `cat` to create a schema definition file, `product_schema.yaml`, using
+   the following contents.
+
+   ```
+   root@gridd-alpha:/# cat > product_schema.yaml
```
dplumb94, line 280: It looks like this is actually a product yaml file. 

Produces this error when using CLI:
```
root@gridd-alpha:/# grid -k alpha-agent \
>         schema create product_schema.yaml
InvalidYamlError: Missing `number_exponent` field for property definition with type NUMBER.
```
Also we should probably put something that looks close to a GS1 schema here (even though its not officially defined..)
Here's an example of a Product that follows the schema that was informally discussed previously:

```cat > product.yaml
   - product_type: "GS1"
     product_id: "013600000929"
     owner: "314156"
     properties:
       - name: "product_name"
         data_type: "STRING"
         string_value: "Truvia 80 ct."
       - name: "image_url"
         data_type: "STRING"
         string_value:
          "https://target.scene7.com/is/image/Target/GUEST_b7a6e983-b391-40a5-ad89-2f906bce5743?fmt=png&wid=1400&qlt=80"
       - name: "brand_name"
         data_type: "STRING"
         string_value: "Truvia"
       - name: "product_description"
         data_type: "STRING"
         string_value: "Truvia Sugar 80CT"
       - name: "gpc"
         data_type: "NUMBER"
         number_value: 30016951
       - name: "net_content"
         data_type: "STRING"
         string_value: "80CT"
       - name: "target_market"
         data_type: "NUMBER"
         number_value: 840
```

so a schema based on that might be a good thing to put in the example.

dplumb94 (changes requested):

vaporos (approved):

dplumb94 (approved):

PR #312, opened 2020-09-03
Update commit errors

dnewh:
Update the commit store errors to use the shared store errors. This helps reduce the amount of repeated code between the store modules.


sdk/src/lib.rs
```diff
@@ -30,6 +30,7 @@ extern crate diesel_migrations;
 #[cfg(feature = "diesel")]
 pub mod database;
 pub mod grid_db;
+pub mod hex;
```
vaporos, line 33: This should not be pub

sdk/src/grid_db/error.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use std::error::Error;
+use std::fmt;
+
+#[cfg(feature = "diesel")]
+use crate::database::error;
+
+/// Represents Store errors
+#[derive(Debug)]
+pub enum StoreError {
```
vaporos, line 23: The errors should probably be specific to each store but simpler (there is no reason for so many variants in the enum).

vaporos (changes requested):

PR #319, opened 2020-09-08
Add location store

dnewh:
This adds the postgres and sqlite store implementations and stubs the in-memory implementation (to make it compile) for the location store following the new pattern. This does not currently integrate the new store into the daemon but that integration will come in a later PR.

### DIFF
--- a/sdk/src/grid_db/locations/store/diesel/mod.rs
+++ b/sdk/src/grid_db/locations/store/diesel/mod.rs
@@ -1,0 +1,274 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod models;
+mod operations;
+pub(in crate::grid_db) mod schema;
+
+use diesel::r2d2::{ConnectionManager, Pool};
+
+use super::diesel::models::{
+    LocationAttributeModel, LocationModel, NewLocationAttributeModel, NewLocationModel,
+};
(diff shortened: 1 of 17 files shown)


dnewh:
This is built on top of PR #310

sdk/src/store/memory.rs
```diff
@@ -43,4 +45,8 @@ impl StoreFactory for MemoryStoreFactory {
     fn get_grid_organization_store(&self) -> Box<dyn OrganizationStore> {
         Box::new(self.grid_organization_store.clone())
     }
+
+    fn get_grid_location_store(&self) -> Box<dyn LocationStore> {
```
dplumb94, line 49: Is there any reason for stubbing out this implementation but implementing it for organizations?

vaporos (approved):

sdk/src/store/memory.rs
dnewh: The organizations and commits stores were done before we decided to not implement the in-memory stores. The memory module will be deleted in a later PR

sdk/src/grid_db/locations/store/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+    pub enum_value: Option<i32>,
+    pub struct_values: Option<Vec<LocationAttribute>>,
+    pub lat_long_value: Option<LatLongValue>,
+    // The indicators of the start and stop for the slowly-changing dimensions.
+    pub start_commit_num: i64,
+    pub end_commit_num: i64,
+    pub service_id: Option<String>,
+}
+
+#[cfg(feature = "diesel")]
```
vaporos, line 63: Diesel implementation should be entirely under diesel:: and shouldn't change the definition of the store API.

vaporos (changes requested):

sdk/src/grid_db/locations/store/mod.rs
dnewh: updated

sdk/src/grid_db/locations/store/diesel/mod.rs
```diff
@@ -0,0 +1,345 @@
...
+        })?)
+        .update_location(
+            location.into(),
+            make_location_attribute_models(&attributes, None),
+            current_commit_num,
+        )
+    }
+}
+
+#[cfg(feature = "diesel")]
```
vaporos, line 189: This is unnecessary (another below this), as the top-level module already has the feature check

vaporos (approved):

dplumb94 (approved):

PR #321, opened 2020-09-10
Fix ConnectionPool bug

dnewh:
This fixes a bug with a method signature passing the incorrect arguments
to a ConnectionPool. This was causing integration tests to fail.

### DIFF
--- a/daemon/src/rest_api/routes/mod.rs
+++ b/daemon/src/rest_api/routes/mod.rs
@@ -280,8 +280,14 @@ mod test {
         }
     }
 
-    fn get_connection_pool() -> ConnectionPool {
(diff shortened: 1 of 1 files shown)


daemon/src/rest_api/routes/mod.rs
```diff
@@ -280,8 +280,14 @@ mod test {
         }
     }
 
-    fn get_connection_pool() -> ConnectionPool {
-        database::create_connection_pool(&DATABASE_URL).expect("Unable to unwrap connection pool")
+    #[cfg(feature = "postgres")]
```
vaporos, line 283: Both of these features will be enabled at the same time.

vaporos (changes requested):

PR #323, opened 2020-09-11
Product Stores

RyanLassigBanks:
Adds ProductStore traits and two implementations. One for Postgresql and one Sqlite.

### DIFF
--- a/sdk/src/grid_db/products/store/diesel/mod.rs
+++ b/sdk/src/grid_db/products/store/diesel/mod.rs
@@ -1,0 +1,301 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub(in crate::grid_db) mod models;
+mod operations;
+pub(in crate::grid_db) mod schema;
+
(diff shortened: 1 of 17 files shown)


sdk/src/grid_db/products/store/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub product_address: String,
+    pub product_namespace: String,
+    pub owner: String,
+    pub start_commit_num: i64,
+    pub end_commit_num: i64,
+    pub service_id: Option<String>,
+    pub properties: Vec<PropertyValue>,
+}
+
+#[cfg(feature = "diesel")]
```
vaporos, line 38: The diesel feature should be entirely under the product::diesel module. We should not be exposing diesel-specific APIs just because diesel is enabled.

vaporos (changes requested):

sdk/src/grid_db/products/store/mod.rs
RyanLassigBanks: Could you elaborate on how you would like this set up? The commit store also uses the diesel feature flag in it's [store/mod.rs](https://github.com/hyperledger/grid/blob/9a8c0926904470242df79da291f978b6258e1833/sdk/src/grid_db/commits/store/mod.rs#L124) so I'm sure what issue is.

sdk/src/grid_db/products/store/diesel/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+use diesel::r2d2::{ConnectionManager, Pool};
+
+use super::{Product, ProductStore, ProductStoreError};
+
+#[derive(Clone)]
+pub struct DieselProductStore<C: diesel::Connection + 'static> {
+    connection_pool: Pool<ConnectionManager<C>>,
+}
+
+#[cfg(feature = "diesel")]
```
vaporos, line 36: This feature guard should be unnecessary here.

vaporos (changes requested):

dplumb94 (dismissed):

vaporos (approved):

dplumb94 (approved):

PR #324, opened 2020-09-11
Add agent store

dnewh:
This adds the postgres and sqlite store implementations and stubs the in-memory implementation (to make it compile) for the agent store following the new pattern. This does not currently integrate the new store into the daemon but that integration will come in a later PR.

### DIFF
--- a/sdk/src/grid_db/agents/store/error.rs
+++ b/sdk/src/grid_db/agents/store/error.rs
@@ -1,0 +1,110 @@
+// Copyright 2018-2020 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::error::Error;
+use std::fmt;
+
+/// Represents AgentStore errors
+#[derive(Debug)]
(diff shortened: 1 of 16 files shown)


sdk/src/grid_db/agents/store/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+pub struct Role {
+    pub public_key: String,
+    pub role_name: String,
+    // The indicators of the start and stop for the slowly-changing dimensions.
+    pub start_commit_num: i64,
+    pub end_commit_num: i64,
+    pub service_id: Option<String>,
+}
+
+#[cfg(feature = "diesel")]
```
vaporos, line 52: Diesel implementation should be under store::diesel and shouldn't change the store API when enabled.

vaporos (changes requested):

sdk/src/grid_db/agents/store/mod.rs
dnewh: updated

sdk/src/grid_db/agents/store/error.rs
```diff
@@ -0,0 +1,137 @@
...
+            AgentStoreError::DuplicateError {
+                context,
+                source: None,
+            } => write!(f, "The agent already exists: {}", context),
+            AgentStoreError::NotFoundError(ref s) => write!(f, "Agent not found: {}", s),
+        }
+    }
+}
+
+#[cfg(feature = "diesel")]
```
RyanLassigBanks, line 115: Should these also be in the diesel module?

RyanLassigBanks (dismissed):

sdk/src/grid_db/agents/store/error.rs
dnewh: updated

RyanLassigBanks (approved):

dplumb94 (approved):

vaporos (approved):

PR #339, opened 2020-09-21
Add shared circuit-template volume

shannynalayna:
Adds a shared-volume to the Splinter example's docker-compose file
amongst the gridd and splinterd containers to store any circuit template
files, so both the Grid daemon and Splinter daemon have access to these
files.

### DIFF
--- a/daemon/packaging/ubuntu/postinst
+++ b/daemon/packaging/ubuntu/postinst
@@ -36,6 +36,12 @@ case "$1" in
         chown -R root:$group /etc/grid
         chmod 775 /etc/grid
         chmod 750 /etc/grid/keys
+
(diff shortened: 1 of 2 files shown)


daemon/packaging/ubuntu/postinst
```diff
@@ -36,6 +36,12 @@ case "$1" in
         chown -R root:$group /etc/grid
         chmod 775 /etc/grid
         chmod 750 /etc/grid/keys
+
```
peterschwarz, line 39: Separate these into two commits - one for the postinst change, and one for the docker change.

shannynalayna: Makes sense, split that commit into two

peterschwarz (approved):

rbuysse (approved):

PR #399, opened 2020-10-26
Have location show command correctly consume location_id

RyanLassigBanks:
Fixes small bug that was causing the location_id being passed to the
show command to not be read.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -238,7 +238,7 @@ fn run() -> Result<(), CliError> {
             )
             (@subcommand show =>
                 (about: "Show location specified by ID argument")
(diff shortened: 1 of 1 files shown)


dnewh (approved):

dplumb94 (changes requested):
I'm still getting an unwrap error here: 

```
root@gridd-alpha:/# grid location list
ID            NAMESPACE OWNER0013600000011 GS1       013600
root@gridd-alpha:/# grid location show 0013600000011
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/src/main.rs:826:47
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

dplumb94 (approved):

PR #400, opened 2020-10-27
Add version headers

dnewh:
This adds protocol version headers to client and UI requests. This is currently not checked but will be in the future so this is a forward-compatibility change.

### DIFF
--- a/cli/src/http.rs
+++ b/cli/src/http.rs
@@ -38,7 +38,11 @@ pub fn submit_batches(
         final_url = format!("{}?service_id={}", final_url, service_id);
     }
     debug!("url {}", final_url);
-    let mut response = client.post(&final_url).body(bytes).send()?;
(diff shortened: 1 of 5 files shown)


cli/src/http.rs
```diff
@@ -40,6 +40,7 @@ pub fn submit_batches(
     debug!("url {}", final_url);
     let batch_link = client
         .post(&final_url)
+        .header("GridProtocolVersion", "0.1")
```
vaporos, line 43: This should be "1" instead of "0.1".

vaporos (changes requested):

cli/src/http.rs
dnewh: updated

vaporos (approved):

dplumb94 (approved):

PR #415, opened 2020-10-28
Fix global CLI options

dnewh:
This moves the key and service-id args one level deeper. These args were previously global to all grid commands but were not relevant to keygen and admin subcommands. Now these args must be specified after the grid subcommand rather than the grid command and are only available on relevant subcommands. This also updates the Grid-on-Splinter example README to reflect these changes.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -86,16 +86,14 @@ fn run() -> Result<(), CliError> {
         (version: VERSION)
         (author: "Contributors to Hyperledger Grid")
         (about: "Command line for Hyperledger Grid")
-        (@arg url: --url +global +takes_value "URL for the REST API")
-        (@arg wait: --wait +global +takes_value "How long to wait for transaction to be committed")
-        (@arg key: -k +global +takes_value "base name for private key file")
         (@arg verbose: -v +multiple +global "Log verbosely")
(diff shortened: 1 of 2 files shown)


dplumb94 (changes requested):
This should include the --wait to only show for creation commands and --url only for commands that have to do with the REST API.

dplumb94 (dismissed):

cli/src/main.rs
```diff
@@ -178,25 +191,32 @@ fn run() -> Result<(), CliError> {
         (@subcommand product =>
             (about: "Create, update, or delete products")
             (@setting SubcommandRequiredElseHelp)
+            (@arg url: --url +takes_value "URL for the REST API")
+            (@arg key: -k +takes_value "base name for private key file")
```
peterschwarz, line 195: Isn't the key command only relevant to the commands that also have `wait`? I.e. both are related to submitting transactions.

dnewh: I believe you also need it for fetching records unlike wait

peterschwarz: That does not look to be the case

dnewh: updated

peterschwarz (approved):

dplumb94 (approved):

PR #418, opened 2020-10-29
Remove quotes from keygen output

dnewh:
This removes the quotation marks from around the file path output in the
`grid keygen` and `grid admin keygen` commands. This is a more standard
output format.

### DIFF
--- a/cli/src/actions/admin.rs
+++ b/cli/src/actions/admin.rs
@@ -99,9 +99,9 @@ pub fn do_keygen(
     let public_key = context.get_public_key(&*private_key)?;
 
     if public_key_path.exists() {
-        info!("Overwriting file: {:?}", public_key_path);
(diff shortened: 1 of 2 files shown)


cli/src/actions/admin.rs
```diff
@@ -99,9 +99,9 @@ pub fn do_keygen(
     let public_key = context.get_public_key(&*private_key)?;
 
     if public_key_path.exists() {
-        info!("Overwriting file: {:?}", public_key_path);
+        info!("Overwriting file: {}", &public_key_path.display());
```
vaporos, line 102: Should be no need for & .display() here, switching from "{:?}" to "{}" should be enough.

vaporos: Nevermind, these are paths.

vaporos (approved):

dplumb94 (approved):

PR #420, opened 2020-11-02
Update product list output

dnewh:
This updates the `grid product list` command's output to display a
subset of the product data in a table format. This improves readability
and enables the use of other command line utilities for searching and
filtering results.

### DIFF
--- a/cli/src/actions/products.rs
+++ b/cli/src/actions/products.rs
@@ -64,6 +64,40 @@ pub struct LatLong {
 }
 
 /**
+ * Prints basic info for products
+ *
+ * products - Products to be printed
+ */
+pub fn display_products_info(products: &[GridProduct]) {
+    // GTINs are always 14 characters long
+    const ID_LENGTH: usize = 14;
+    // Column header "namespace" is longer than the namespace strings in practice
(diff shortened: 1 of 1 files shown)


cli/src/actions/products.rs
```diff
@@ -63,6 +63,26 @@ pub struct LatLong {
     longitude: i64,
 }
 
+/**
+ * Prints basic info for products
+ *
+ * products - Products to be printed
+ */
+pub fn display_products_info(products: &[GridProduct]) {
+    println!("{:<14} {:<9} {:<16}", "ID", "NAMESPACE", "OWNER");
```
vaporos, line 72: Both println!() in this function should use the same format string.

I'd probably define the lengths as consts at the beginning of the function, with some comments "// GTINs are 14 characters", "header is longer than namespace strings", etc. Maybe the format string could even be const since it doesn't change. The owner field being the last one doesn't need the arbitrary 16-width fill.

dnewh: updated

vaporos (approved):

dplumb94 (approved):

jsmitchell (approved):

Issue #411, opened 2020-10-27
Vague error message when no service_id isn't specified 


RyanLassigBanks:
When running grid in splinter mode, if you do not specify the service_id in the CLI you get the following error message

```
grid -k alpha-agent organization create 314156 myorg street --metadata gs1_company_prefixes=9012345000004
Reqwest Error: expected value at line 1 column 1
```

This error should say something about how the service_id was not specified

Issue #407, opened 2020-10-27
Handle "Reqwest Error" errors in the Grid CLI more elegantly


dplumb94:
root@gridd-alpha:/# grid -k alpha-agent --wait 300 location create 1234 --namespace GS1
Reqwest Error: invalid type: string "Could not find schema with name: GS1", expected struct GridSchemaSlice at line 1 column 38
 ^ It appears that in this case we are not checking for the error. We should be handling these HTTP errors more elegantly.

PR #425, opened 2020-11-06
Grid CLI error messages fixes

RyanLassigBanks:
Closes [issue 407](https://github.com/hyperledger/grid/issues/407) and [issue 411](https://github.com/hyperledger/grid/issues/411)

### DIFF
--- a/cli/src/actions/locations.rs
+++ b/cli/src/actions/locations.rs
@@ -94,10 +94,13 @@ pub fn do_list_locations(url: &str, service_id: Option<&str>) -> Result<(), CliE
     if let Some(service_id) = service_id {
         final_url = format!("{}?service_id={}", final_url, service_id);
     }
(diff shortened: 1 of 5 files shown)


vaporos (approved):

dplumb94 (approved):
Big improvement, lgtm. I did notice one thing:
```
MacBook-Pro:debug dplumb$ ./grid -k alpha-agent --wait 300 location create 1234 --namespace GS1
Circuit ID is not present, but grid is running in splinter mode
```

It would be nice if the CLI could tell you what you need to do in cases like this, something like:

`You must the GRID_SERVICE_ID environment variable or include the --service-id option when running in splinter mode`

However since this error is only coming from the CLI it seems hard to do that in a general way. We would probably have to add some pre-checks in the CLI to do that. Something to think about for the future.

PR #430, opened 2020-11-10
Refactor SDK errors

dnewh:
This refactors the Grid SDK to implement and use new top-level errors similar to the implementation in `libsplinter`. This improves consistency and ease-of-use in error handling within Grid.


sdk/src/database/error.rs
```diff
@@ -32,13 +32,17 @@ impl fmt::Display for ConnectionError {
         write!(f, "Unable to connect to database: {}", self.context)
     }
 }
+use crate::error::{
```
ltseeley: I'd split these changes that actually use the error into a separate commit

sdk/src/database/error.rs
```diff
@@ -76,10 +80,9 @@ impl fmt::Display for DatabaseError {
 
 impl From<diesel::ConnectionError> for DatabaseError {
```
peterschwarz: Given we use a connection pool, what error would we need to map to report this if a query failed due to the connection to the db dropping?  This can be done in a future PR.

sdk/src/database/error.rs
dnewh: Updated

sdk/src/database/error.rs
```diff
@@ -88,8 +91,25 @@ impl From<diesel_migrations::RunMigrationsError> for DatabaseError {
 
 impl From<diesel::result::Error> for DatabaseError {
     fn from(err: diesel::result::Error) -> Self {
-        DatabaseError::QueryError(Box::new(err))
         match err {
+            diesel::result::Error::DatabaseError(
```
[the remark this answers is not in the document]
dnewh: This entire db module actually goes away in this PR. I updated the commits to better reflect this change

peterschwarz (approved):

dplumb94 (approved):

PR #441, opened 2020-11-19
Add feature guards to CLI

RyanLassigBanks:
The grid cli feature set is now guarded by the following features:

- pike
- product
- schema
- location

When these features are enabled the corresponding set of cli commands
will be enabled. Pike, product, schema, and location are considered
stable features and will be included if the cli is compiled with the
stable flag.

When only location and product are enabled, pike and schema will be
included as well because those features are required to properly use
product and location. Similarly, if schema is enabled pike will also be
included for the same reason.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -49,10 +49,14 @@ diesel_migrations = "1.4"
 [features]
 default = ["sawtooth"]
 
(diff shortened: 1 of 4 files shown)


dnewh (approved):
tested, lgtm

dplumb94 (approved):

PR #446, opened 2020-11-30
Update keygen command

dnewh:
This updates the `grid keygen` command to work similarly to its Splinter
counterpart. This change adds a `--system` argument that causes the
action to generate the keys in a system defined directory.


RyanLassigBanks (approved):
Tested locally

RyanLassigBanks (approved):

dplumb94 (approved):

PR #478, opened 2021-01-11
Add Workflow experimental feature

RyanLassigBanks:
Adds the experimental feature for workflows to the SDK

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -66,7 +66,8 @@ experimental = [
     "stable",
     # The following features are experimental:
     "database",
-    "track-and-trace"
+    "track-and-trace",
+    "workflow"
 ]
 
 database = [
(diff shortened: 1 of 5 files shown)


sdk/src/workflows.rs
```diff
@@ -0,0 +1,319 @@
+// Copyright 2018-2021 Cargill Incorporated
```
vaporos, line 1: It would probably be good to start this crate as a directory, minimally broken up into modules for workflow, subworkflow, and state.

RyanLassigBanks: Are you sure that's necessary at this point? The module is pretty small and all of the structs inside it are related to one another

vaporos (approved):

vaporos (approved):

dplumb94 (approved):

PR #507, opened 2021-01-28
Add justfile

vaporos:
Adds a justfile. Fixes many feature guards, linting errors, etc. No intentional functional changes beyond new features being defined.

This branch passes "just build" and "just lint". It does not pass "just test", since tests need adjustment to be run without additional setup.

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -31,10 +31,16 @@ impl GridConfig {
         &self.endpoint
     }
 
+    // dead_code allowed because this isn't used when splinter and sawtooth
+    // features are both off, as is the case with --no-default-features
+    #[allow(dead_code)]
     pub fn rest_api_endpoint(&self) -> &str {
         &self.rest_api_endpoint
     }
 
+    // dead_code allowed because this isn't used when splinter and sawtooth
+    // features are both off, as is the case with --no-default-features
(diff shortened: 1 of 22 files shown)


daemon/src/config.rs
```diff
@@ -31,10 +31,16 @@ impl GridConfig {
...
+    // dead_code allowed because this isn't used when splinter and sawtooth
+    // features are both off, as is the case with --no-default-features
+    #[allow(dead_code)]
     pub fn rest_api_endpoint(&self) -> &str {
         &self.rest_api_endpoint
     }
 
+    // dead_code allowed because this isn't used when splinter and sawtooth
+    // features are both off, as is the case with --no-default-features
+    #[allow(dead_code)]
```
peterschwarz, line 43: Could this be `feature = "database"`?

vaporos: When I started down that path, it felt like the logical conclusion wasn't matching the intent -- i.e. these options are not actually optional. Like, if we delete this function, then we would have to put dead_code on the struct member. If we remove the struct member, then we need to handle this in the builder. And that's what doesn't feel correct to make conditional.

peterschwarz (approved):

agunde406 (approved):

PR #510, opened 2021-01-29
Include record owner in has_permission check

dplumb94:
Updates the has_permission check to include `record_owner`, which is the Pike ID of the org that owns the record that we are checking permissions for.  This also adds a check to make sure the agent actually belongs to that org and removes it from the smart contract. This is part of an effort to move common code from the contracts to the SDK.

### DIFF
--- a/contracts/product/src/state.rs
+++ b/contracts/product/src/state.rs
@@ -23,11 +23,10 @@ cfg_if! {
 }
 
 use grid_sdk::{
-    agents::addressing::compute_agent_address,
     organizations::addressing::compute_organization_address,
     products::addressing::compute_gs1_product_address,
     protocol::{
-        pike::state::{Agent, AgentList, Organization, OrganizationList},
+        pike::state::{Organization, OrganizationList},
         product::state::{Product, ProductList, ProductListBuilder},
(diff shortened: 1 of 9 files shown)


dnewh (dismissed):

dnewh (approved):

contracts/product/src/state.rs
```diff
@@ -173,34 +172,6 @@ impl<'a> ProductState<'a> {
         Ok(())
     }
 
-    /// Gets a Pike Agent. Handles retrieving the correct agent from an AgentList.
-    pub fn get_agent(&self, public_key: &str) -> Result<Option<Agent>, ApplyError> {
```
peterschwarz, line 177: Likewise, it would be good to explain why these are removed.  If they are removed based on no longer being used, move the removal to a separate commit.

peterschwarz (changes requested):

contracts/product/src/state.rs
dplumb94: Updated the commit messages to explain the removal of unit tests

peterschwarz (approved):

dnewh (approved):

PR #516, opened 2021-02-04
Consolidate pike db modules

dnewh:
This consolidates the agents and organizations db modules into one "pike" module. The functionality doesn't change at all but now all the pike operations are in one place. This simplifies some code and also makes it clearer that these modules are for the Pike smart contract. This also removes the unused memory store implementation for organizations. This was the last module with a memory backed store implementation so this module was removed from the library.


RyanLassigBanks (dismissed):

sdk/src/store/mod.rs
```diff
@@ -85,6 +84,10 @@ pub fn create_store_factory(
             })?;
             Ok(Box::new(sqlite::SqliteStoreFactory::new(pool)))
         }
+        #[cfg(not(feature = "sqlite"))]
```
vaporos, line 87: Why would this be special for sqlite and not postgres? But also, isn't the match statement complete without this block?

The error message in the block doesn't give any clues about what actually happened.

vaporos (changes requested):

sdk/src/store/mod.rs
dnewh: updated to check for both features and to have a more descriptive error message

dnewh: Removing the block completely was causing errors for at least some combinations of features

RyanLassigBanks (approved):

vaporos (approved):

PR #517, opened 2021-02-04
Stabilize database and splinter in CLI and daemon for 0.1

dplumb94:


vaporos (approved):

cli/Cargo.toml
```diff
@@ -54,11 +54,12 @@ default = [
     "sawtooth",
     "database",
     "postgres",
+    "sqlite",
 ]
 
-stable = ["location", "pike", "product", "schema", "default"]
+stable = ["location", "pike", "product", "schema",  "postgres", "sqlite", "splinter", "default"]
```
peterschwarz, line 60: This adds postgres to stable, but doesn't seem to be moving it from anywhere. Was that an oversight?

dplumb94: I believe so, its a bit confusing how it was set up. `stable` was including `default` which has `postgres` in it. Here im adding both to stable and sqlite to default. I'm not sure if this is the right thing to do tbh.

vaporos (approved):

rbuysse (dismissed):

peterschwarz (approved):

rbuysse (approved):

peterschwarz (approved):

PR #520, opened 2021-02-05
Grid daemon Pagination

RyanLassigBanks:
Adds pagination to all routes that list multiple elements. Now, instead of returning a list, these endpoints return objects that contain a fixed number of elements and pagination data
```
{
   "data" : [ ... ]
   "paging": {
      "current": "/agent?offset0&limit=10",
      "offset": 0,
      "limit": 10,
      "total": 1000,
      "prev": "/agent?offset=0&limit=10",
      "next": "/agent?offset=10&limit=10",
      "last": "/agent?offset100&limit=10"
    }
}
  ```

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -33,7 +33,12 @@ extern crate log;
 ))]
 #[macro_use]
 extern crate serde_json;
-#[cfg(feature = "splinter-support")]
+#[cfg(any(
+    feature = "pike",
+    feature = "schema",
+    feature = "product",
+    feature = "location"
+))]
 #[macro_use]
 extern crate serde;
 
--- a/daemon/src/rest_api/routes/agents.rs
+++ b/daemon/src/rest_api/routes/agents.rs
@@ -15,7 +15,9 @@
 use std::{convert::TryFrom, str::FromStr};
 
 use crate::rest_api::{
-    error::RestApiResponseError, routes::DbExecutor, AcceptServiceIdParam, AppState, QueryServiceId,
+    error::RestApiResponseError,
+    routes::{paging::Paging, DbExecutor},
+    AcceptServiceIdParam, AppState, QueryPaging, QueryServiceId,
 };
 
 use actix::{Handler, Message, SyncContext};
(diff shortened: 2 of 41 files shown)

### CODE REVIEW

File: daemon/src/rest_api/routes/agents.rs [hunk 1]
```diff
@@ -36,6 +38,12 @@ pub struct AgentSlice {
     pub service_id: Option<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AgentListSlice {
```
peterschwarz, line 42: This could probably be a generic type `PagedData<T>` or `PagedSlice<T>`, as the only thing that varies on this is the type in the data vector.

daemon/src/rest_api/routes/agents.rs
```diff
@@ -63,33 +71,47 @@ impl TryFrom<Agent> for AgentSlice {
 
 struct ListAgents {
     service_id: Option<String>,
+    offset: i64,
+    limit: i64,
```
peterschwarz, line 75: I wonder if we should enforce a max through the type here - no more than u8? or u16?

daemon/src/rest_api/routes/agents.rs
```diff
@@ -63,33 +71,47 @@ impl TryFrom<Agent> for AgentSlice {
 
 struct ListAgents {
     service_id: Option<String>,
+    offset: i64,
```
peterschwarz, line 74: Can this be negative?  If not, we should probably use the usigned type at this level.

peterschwarz (commented):
The comments apply to all the commits, as the same items occur in every one.

peterschwarz (approved):
Approving, taking the questions to a design session for a future PR.

daemon/src/main.rs
```diff
@@ -33,7 +33,12 @@ extern crate log;
 ))]
 #[macro_use]
 extern crate serde_json;
-#[cfg(feature = "splinter-support")]
+#[cfg(any(
```
vaporos, line 36: This should probably be #[cfg(feature = "serde")]

daemon/src/rest_api/routes/agents.rs
vaporos: I'd be ok with this if it is documented as a best practice, since it deviates from our general position that endpoints should not re-use datatypes (what differs here is perhaps that they are solely used for REST APIs, so we could version them as well without impacting the rest of the library).

sdk/src/paging.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018-2020 Cargill Incorporated
```
vaporos, line 1: This should go in the crate::rest_api::paging

vaporos (approved):

PR #521, opened 2021-02-08
Update Pike db module tables

dnewh:
This updates the Pike DB module with updated/new tables in preparation for the Pike 2.0 update. This adds some tables that are not used quite yet but are defined in the Identity RFC. This also includes the associated migrations for postgres and sqlite backends and relevant daemon/cli updates.

Issue #512, opened 2021-02-02
Grid SDK - Update Pike DB table structure


dnewh:
Update the Pike Database table structure to the new design. This may include adding some additional tables that are not used yet to prepare for the Pike 2.0 update.

PR #530, opened 2021-02-15
Add batch-store to to_hex feature guard

RyanLassigBanks:
Batch store uses to_hex to serialize batches, so it needs to be added to
its feature guard.


RyanLassigBanks:
Handled in a [#512](https://github.com/hyperledger/grid/pull/521)

PR #531, opened 2021-02-15
Add Pike 2.0 create and update

dnewh:
This adds functionality for creating and updating roles along with other Pike 2.0 updates. This includes significant updates to the Pike db store/SDE, the Pike smart contract and any smart contract that uses Pike, the Grid REST API, and the Grid CLI.

# Testing

- Follow the Grid on Splinter walkthrough instructions through creating an organization.
- Use the `grid role create` command to create a role with the permissions `schema::can-create-schema,product::can-create-product` and set it as active with the `--active` flag
- use the `grid agent update` command to add the new role to the agent being sure not to forget to specify the `admin` role as well.
- create a product schema and product


sdk/protos/pike_payload.proto
```diff
@@ -52,16 +64,51 @@ message UpdateAgentAction {
...
+  string org_id = 1;
+  string public_key = 2;
+}
+
 message CreateOrganizationAction {
   string id = 1;
   string name = 2;
-  string address = 3;
-  repeated KeyValueEntry metadata = 4;
+  repeated AlternateID alternate_ids = 4;
```
dplumb94, line 75: skipped a number here

sdk/protos/pike_state.proto
```diff
@@ -27,18 +27,44 @@ message AgentList {
...
+  repeated string locations = 3;
+  repeated AlternateID alternate_id = 4;
+  repeated KeyValueEntry metadata = 5;
 }
 
 message OrganizationList {
   repeated Organization organizations = 1;
 }
+
+message Role {
```
dplumb94, line 42: These can be numbered in order since we arent worried about backward compatibility

sdk/src/permissions.rs
```diff
@@ -147,16 +252,32 @@ mod tests {
...
 
     const PUBLIC_KEY: &str = "test_public_key";
     const ORG_ID: &str = "test_org";
     const WRONG_ORG_ID: &str = "test_wrong_org";
 
+    const PERM_CAN_DRIVE: &str = "tankops::can-drive";
+    const PERM_CAN_TURN_TURRET: &str = "tankops::can-turn-turret";
+    const PERM_CAN_FIRE: &str = "tankops::can-fire";
+
+    const ROLE_ALPHA_INSPECTOR: &str = "Inspector";
```
dplumb94, line 276: It looks like the prefixes were removed from these consts. I would recommend either not using the variables for these tests or keeping the prefix if using them as external roles. I think that just removing the prefixes makes the tests a bit confusing. Also in some places the org ID consts are not being used.

sdk/src/permissions.rs
```diff
@@ -138,6 +207,42 @@ impl<'a> PermissionChecker<'a> {
...
+        let (name, org_id) = match org_id {
+            Some(org_id) => (name, org_id),
+            None => {
+                if name.contains('.') {
+                    let t: Vec<&str> = name.split('.').collect();
+                    let org_id = t[0];
+                    let name = t[1];
+                    (name, org_id)
+                } else {
+                    return Err(PermissionCheckerError::InvalidRole("External roles need to be prefixed with their org ID. Format: <org_id>.<role_name>".to_string()));
```
dplumb94, line 225: Hmm this is interesting, so for `inherits_from` we are still using the dot-separated role names. Maybe that's fine but it makes me think it should be a struct or something. Not blocking for this but probably needs some discussion before we release a version of Grid with this in it.

dnewh: it is only dot separated for external roles, similar to how adding an external role to an agent would work. Internal roles can still be referenced just by the role name

dplumb94 (changes requested):
Getting an error trying to view roles via CLI:

```
root@gridd-alpha:/# grid role create --permissions schema::can-create-schema,product::can-create-product crgl ProductOwner 'role to create products' -v
...
root@gridd-alpha:/# grid role show crgl ProductOwner
missing field `allowed_orgs` at line 1 column 223
```

sdk/src/permissions.rs
dnewh: I disagree with this. I think having the consts as the role names potentially makes the tests easier to maintain down the line and having them without the prefix and adding that on for external roles reflects how people would actually use that.

dplumb94 (approved):
Tested, works. I kind of feel like we should wrap this stuff in a feature. At the very least we need to go through the stabilization process for it before the next grid release.

RyanLassigBanks (approved):

PR #572, opened 2021-03-15
Add alternate ID support

dnewh:
This adds alternate ID support to the Pike, Product, and Location smart contracts. This replaces the functionality of the special case of organization metadata that was being checked for GS1 company prefixes with a more flexible field on the Organization object. This field can be used for any number of alternate identifiers for an organization.

## Testing

- Follow grid walkthrough directions to set up a circuit
- create an organization and add the `gs1_company_prefixes` ID with the `--alternate-ids` flag
- create a role to create schemas, products, and locations
- Add the role to the agent
- create schemas for products and locations
- create a product and create a location

### DIFF
--- a/contracts/location/src/handler.rs
+++ b/contracts/location/src/handler.rs
@@ -171,16 +171,16 @@ fn create_location(
     )?;
 
     // check if organization has gln in gs1_company_prefix metadata
-    let mut has_gs1_prefix = false;
-    for metadata in organization.metadata() {
-        if metadata.key() == "gs1_company_prefixes"
-            && payload.location_id().contains(metadata.value())
+    let mut has_gs1_alt_id = false;
+    for alternate_id in organization.alternate_ids() {
+        if alternate_id.id_type() == "gs1_company_prefix"
+            && payload.location_id().contains(alternate_id.id())
         {
-            has_gs1_prefix = true;
+            has_gs1_alt_id = true;
         }
     }
 
-    if !has_gs1_prefix {
+    if !has_gs1_alt_id {
         return Err(ApplyError::InvalidTransaction(format!(
             "Organization {} does not have the correct gs1 prefix",
             organization.org_id()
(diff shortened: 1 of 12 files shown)


RyanLassigBanks (approved):

contracts/location/src/handler.rs
```diff
@@ -171,18 +171,18 @@ fn create_location(
     )?;
 
     // check if organization has gln in gs1_company_prefix metadata
-    let mut has_gs1_prefix = false;
-    for metadata in organization.metadata() {
-        if metadata.key() == "gs1_company_prefixes"
-            && payload.location_id().contains(metadata.value())
+    let mut has_gs1_alt_id = false;
+    for alternate_id in organization.alternate_ids() {
+        if alternate_id.id_type() == "gs1_company_prefixes"
```
dplumb94, line 176: This should really be "gs1_company_prefix". Each org may have several of these but the key of the ID itself should be "gs1_company_prefix"

contracts/location/src/handler.rs
```diff
@@ -171,18 +171,18 @@ fn create_location(
...
-            has_gs1_prefix = true;
+            has_gs1_alt_id = true;
         }
     }
 
-    if !has_gs1_prefix {
+    if !has_gs1_alt_id {
         return Err(ApplyError::InvalidTransaction(format!(
-            "Organization {} does not have the correct gs1 prefix",
+            "Organization {} does not have the correct gs1 alternate ID",
```
dplumb94, line 185: This can still say "prefix" because the actual ID type is a GS1 company prefix.

contracts/product/src/handler.rs
```diff
@@ -136,24 +136,24 @@ impl ProductTransactionHandler {
...
-                .find(|kv| kv.key() == "gs1_company_prefixes")
+                .find(|p| p.id_type() == "gs1_company_prefixes")
             {
-                Some(gs1_company_prefix_kv) => gs1_company_prefix_kv,
+                Some(gs1_company_prefix) => gs1_company_prefix,
                 None => {
                     return Err(ApplyError::InvalidTransaction(format!(
-                    "The agents organization does not have the gs1_company_prefixes key in its metadata: {:?}",
-                    org.metadata()
+                    "The agents organization does not have the gs1_company_prefixes alternate ID: {:?}",
```
dplumb94, line 147: same as location comment

dplumb94 (changes requested):

RyanLassigBanks (approved):

dplumb94 (approved):

PR #583, opened 2021-03-24
Create `Client` trait

isabeltomb:
Create a `Client` trait in a `client` module in sdk/src/ that can be implemented with different backends. The initial implementation of this client trait contains method definitions for purchase order related interactions. This client is guarded by the `client` experimental feature.

### DIFF
--- a/sdk/src/client/mod.rs
+++ b/sdk/src/client/mod.rs
@@ -1,0 +1,149 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


sdk/src/purchase_order/client/mod.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2018-2021 Cargill Incorporated
```
vaporos, line 1: Move to sdk/src/client/mod.rs and rename PurchaseOrderClient to Client.

isabeltomb: Fixed

sdk/src/client/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use crate::error::InternalError;
+
+pub trait Client {
+    /// Adds the given `PurchaseOrder`.
+    ///
+    /// # Arguments
+    ///
+    /// * `purchase_order` - The `PurchaseOrder` to be added
+    fn add_purchase_order(&self, purchase_order: &PurchaseOrder) -> Result<(), InternalError>;
```
peterschwarz, line 29: These should all be guarded by the "purchase-order" feature

peterschwarz (changes requested):

sdk/src/client/mod.rs
isabeltomb: Fixed

vaporos (approved):

peterschwarz (approved):

PR #596, opened 2021-03-25
Add reqwest stub

isabeltomb:
Update the Client trait to have `add_purchase_order_version` `update_purchase_order_version` and `add_purchase_order_revision` method definitions.

Add an outline for the reqwest based implementation of the `Client` trait. All methods are unimplemented. The `reqwest` module is guarded by the experimental feature `client-reqwest`

Fix the `PurchaseOrderRevision` and `PurchaseOrderVersion` structs to match protos. Update the `PurchaseOrderRevision` struct to have `revision_id` instead of `revision_number` and update the `PurchaseOrderVersion` to have `current_revision_id` instead of `current_revision_number`

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -80,6 +80,7 @@ experimental = [
     # The following features are experimental:
     "batch-store",
     "client",
+    "client-reqwest",
     "postgres",
     "pike-rest-api",
     "purchase-order",
@@ -92,6 +93,7 @@ experimental = [
 ]
 
 client = []
+client-reqwest = ["client", "reqwest"]
 location = ["pike", "schema"]
 pike = []
 pike-rest-api = ["pike", "serde_json", "rest-api-resources"]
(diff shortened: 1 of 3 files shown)


sdk/src/client/reqwest.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl ReqwestClient {
+    pub fn _new(url: String) -> Self {
+        ReqwestClient { url }
+    }
+}
+
+impl Client for ReqwestClient {
+    /// Adds the given `PurchaseOrder`.
+    #[cfg(feature = "purchase-order")]
```
shannynalayna, line 40: Do we need `add` and `update` methods for po versions and revisions here?

isabeltomb: I believe so, looking at the RFC I only see information about being able to update versions but not revisions so I added a commit to add `add_purchase_order_version` `update_purchase_order_version` and `add_purchase_order_revision` to the client trait

sdk/Cargo.toml
```diff
@@ -92,6 +93,7 @@ experimental = [
 ]
 
 client = []
+client-reqwest = ["client", "reqwest"]
```
shannynalayna, line 96: Probably want to separate adding this feature into a separate commit

isabeltomb: Moved this to its own commit

shannynalayna (approved):

dplumb94 (approved):

PR #601, opened 2021-03-25
Fix lint issues

dnewh:
This resolves a couple lint issues introduces with the latest version of
Rust.

### DIFF
--- a/sdk/src/purchase_order/store/diesel/mod.rs
+++ b/sdk/src/purchase_order/store/diesel/mod.rs
@@ -1,0 +1,13 @@
+// Copyright 2018-2021 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


peterschwarz (approved):

RyanLassigBanks (changes requested):
Clippy looks good, but there's some cargo fmt lint that needs to be fixed

sdk/src/purchase_order/store/diesel/mod.rs
```diff
@@ -0,0 +1 @@
+
```
agunde406, line 1: This looks like it was accidentally included.

dnewh: This came from cargo fmt

vaporos (approved):

RyanLassigBanks (approved):
Clippy looks good, but there's some cargo fmt lint that needs to be fixed

agunde406 (approved):

PR #610, opened 2021-03-26
Fix serialization issues

peterschwarz:
This PR fixes several serialization issues.

### DIFF
--- a/cli/src/actions/schemas.rs
+++ b/cli/src/actions/schemas.rs
@@ -446,9 +446,9 @@ mod test {
     use std::panic;
     use std::thread;
 
-    static LIGHTBULB_YAML_EXAMPLE: &[u8; 816] = br##"- name: "Lightbulb"
+    static LIGHTBULB_YAML_EXAMPLE: &[u8; 818] = br##"- name: "Lightbulb"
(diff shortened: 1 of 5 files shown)


vaporos (approved):

jsmitchell (approved):

vaporos (approved):

jsmitchell (approved):

cli/src/actions/schemas.rs
```diff
@@ -446,9 +446,9 @@ mod test {
     use std::panic;
     use std::thread;
 
-    static LIGHTBULB_YAML_EXAMPLE: &[u8; 816] = br##"- name: "Lightbulb"
+    static LIGHTBULB_YAML_EXAMPLE: &[u8; 818] = br##"- name: "Lightbulb"
```
dplumb94, line 449: Why do the numbers change in these examples?

peterschwarz: Two characters were added to the array, and this is the exact size.

dplumb94: Nvm I see now, diff was confusing

dplumb94 (approved):

PR #614, opened 2021-03-26
Implement `post_batches`

isabeltomb:
- Update the method definition of `post_batches` in the `Client` trait.
- Implement the `post_batches` method in the reqwest based implementation of the `Client` trait.


RyanLassigBanks (approved):
I'm not sure why the `Client` implementation is called `ReqwestClient`, or why calling out the underlying http library is interesting.

dplumb94 (approved):

PR #629, opened 2021-04-02
Implement create product using XML data

dplumb94:
This is the first PR to implement XML support for product. The code was getting a bit complex so I am submitting this one first to keep it incremental. Remaining tasks include:

- Implementing update
- Validation
- Fetching alternate ID to fill in the owner

To test: 

1. start up the `examples/splinter/docker-compose.yaml` network and create a Grid circuit between alpha and beta
2. Set up agents and organizations as follows:

```
grid keygen alpha-agent

export GRID_SERVICE_ID=<circuit ID>::gsAA

grid organization create crgl Cargill --metadata gs1_company_prefixes=0
grid role create crgl productowner --permissions schema::can-create-schema,product::can-create-product --active
grid agent update crgl $(cat ~/.grid/keys/alpha-agent.pub) --active --role productowner --role admin
```
3. submit a transaction to create a schema with the new `xsd_schema.yaml` file. If you are executing this from the gridd-alpha container you might need to copy it in. 
`$ grid schema create xsd-schema.yaml`
4. Create a product using some example XML data that conforms to the GridTradeItems XSD.
`$ grid product create -f yourproduct.yaml`

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -70,6 +70,7 @@ experimental = [
     # The experimental feature extends stable:
     "stable",
     # The following features are experimental:
+    "product-gdsn",
     "purchase-order",
     "splinter",
     "sqlite",
@@ -87,6 +88,9 @@ admin-keygen = []
 location = ["pike", "schema"]
 pike = []
 product = ["pike", "schema"]
+product-gdsn = [
+    "grid-sdk/product-gdsn"
+]
 purchase-order = []
 sawtooth = []
 schema = ["pike"]
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -718,6 +718,74 @@ fn run() -> Result<(), CliError> {
     {
         use clap::{Arg, SubCommand};
 
+        #[allow(unused_mut)]
+        let mut create_product_subcmd = SubCommand::with_name("create")
+            .about("Create a product")
+            .arg(
+                Arg::with_name("product_id")
+                    .conflicts_with("file")
+                    .takes_value(true)
+                    .required_unless_one(&["file", "xml"])
+                    .help("Unique ID for product"),
+            )
+            .arg(
+                Arg::with_name("product_namespace")
+                    .long("namespace")
+                    .takes_value(true)
+                    .conflicts_with("file")
+                    .conflicts_with("xml")
+                    .help("Product namespace (example: GS1)"),
+            )
+            .arg(
+                Arg::with_name("owner")
+                    .long("owner")
+                    .takes_value(true)
+                    .conflicts_with("file")
+                    .help("Pike organization ID"),
+            )
+            .arg(
+                Arg::with_name("property")
+                    .long("property")
+                    .use_delimiter(true)
+                    .takes_value(true)
+                    .multiple(true)
+                    .conflicts_with("file")
+                    .help("Key value pair specifying a product property formatted as key=value"),
+            )
+            .arg(
+                Arg::with_name("file")
+                    .long("file")
+                    .short("f")
+                    .takes_value(true)
+                    .conflicts_with("xml")
+                    .help("Path to yaml file containing a list of products"),
+            )
+            .arg(
+                Arg::with_name("key")
+                    .long("key")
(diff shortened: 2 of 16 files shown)


cli/src/main.rs
```diff
@@ -1946,6 +1956,23 @@ fn run() -> Result<(), CliError> {
                 .or_else(|| env::var(GRID_SERVICE_ID).ok());
 
             match m.subcommand() {
+                #[cfg(feature = "xml-support")]
+                ("create", Some(m)) if m.is_present("xml") => {
+                    let key = m
```
vaporos, line 1961: We should create a Grid CLI variant of https://github.com/Cargill/splinter/blob/main/cli/src/signing.rs so these key options work properly.

vaporos (changes requested):

RyanLassigBanks (approved):
Tested locally

cli/src/main.rs
```diff
@@ -766,8 +767,17 @@ fn run() -> Result<(), CliError> {
                                 .long("file")
                                 .short("f")
                                 .takes_value(true)
+                                .conflicts_with("xml")
                                 .help("Path to yaml file containing a list of products"),
                         )
+                        .arg(
+                            Arg::with_name("xml")
```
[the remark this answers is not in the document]
dplumb94: I think another option here would be to name the argument `--gdsn-3-1` (which could correspond with a more descriptive property name of the resulting Grid product definition). I think if we make it completely dynamic we could end up with some strange edge cases with duplicate GTIN product definitions being submitted etc. The complexity in solving that might not be proportional to the actual usefulness of the feature.

cli/src/main.rs
dplumb94: What isnt working properly? I can create a story for that

sdk/Cargo.toml
```diff
@@ -89,7 +89,8 @@ experimental = [
     "sawtooth-compat",
     "sqlite",
     "track-and-trace",
-    "workflow"
+    "workflow",
+    "product-gdsn"
```
vaporos, line 96: This should be in alphabetical order after pike-rest-api

sdk/Cargo.toml
```diff
@@ -114,3 +115,6 @@ sawtooth-compat = [
 ]
 sqlite = ["diesel/sqlite", "diesel_migrations"]
 workflow = []
+product-gdsn = [
```
vaporos, line 121: These should all be in alphabetical order as well (all features after experimental (experimental being special so separate)).

cli/src/main.rs
```diff
@@ -776,8 +777,17 @@ fn run() -> Result<(), CliError> {
                                 .long("file")
                                 .short("f")
                                 .takes_value(true)
+                                .conflicts_with("xml")
                                 .help("Path to yaml file containing a list of products"),
                         )
+                        .arg(
```
vaporos, line 783: This should be behind a feature

cli/src/main.rs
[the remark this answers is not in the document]
vaporos: "--gdsn-3-1" would work but its less user-friendly as it should be easy to determine which type of file it is at runtime, with just --file arguments.

cli/src/main.rs
vaporos: The first is the approach to looking up and handling the key argument, which should follow what is done in the splinter CLI, using code similar to that linked above; the second piece is creating a signer early and not passing the key into functions (pass the signer instead). splinter CLI has many examples of this.

vaporos (changes requested):

cli/Cargo.toml
```diff
@@ -70,9 +70,10 @@ experimental = [
     # The experimental feature extends stable:
     "stable",
     # The following features are experimental:
+    "product-gdsn",
     "purchase-order",
     "splinter",
-    "sqlite",
+    "sqlite"
```
vaporos, line 76: Add the , back to this line pls

vaporos (approved):

dnewh (approved):

PR #635, opened 2021-04-06
Fix UI sidebar

dnewh:
This fixes the sidebar for the grid UI to correctly show icons for the
different saplings. This change depends on the PR to splinter-canopyjs:
https://github.com/Cargill/splinter-canopyjs/pull/21.

### DIFF
--- a/ui/grid-ui/public/index.html
+++ b/ui/grid-ui/public/index.html
@@ -6,6 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <meta name="theme-color" content="#000000" />
   <meta name="description" content="UI for Grid" />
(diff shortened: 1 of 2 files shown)


dplumb94 (changes requested):
Instead of this and the related change in the Splinter admin UI, we can use material icon that goes with each sapling. That way we can ensure that all of the sapling icons have a similar look to them.

dnewh:
> Instead of this and the related change in the Splinter admin UI, we can use material icon that goes with each sapling. That way we can ensure that all of the sapling icons have a similar look to them.

This is still an option. This is just additional functionality

dplumb94 (approved):

RyanLassigBanks (approved):

PR #638, opened 2021-04-13
Refactor grid daemon splinter module

RyanLassigBanks:
Various updates to address comments relating to `splinter-support` feature stabilization


dplumb94 (dismissed):

daemon/src/splinter/event/mod.rs
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Cargill Incorporated
+ * Copyright 2021 Cargill Incorporated
```
vaporos, line 2: Was 2019-2021 intended here?

sdk/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub mod batches;
 pub mod client;
 pub mod commits;
 pub mod error;
-mod hex;
+pub mod hex;
```
vaporos, line 43: Grid SDK is not a hex library, and should not export hex conversion functions.

Because the use case here is "hex::to_hex(signer.public_key());", this can be fixed with converting it to cylinder which allows signer.public_key().as_hex().

vaporos (changes requested):

sdk/src/lib.rs
RyanLassigBanks: Dropped this commit

vaporos (approved):

dplumb94 (approved):

PR #639, opened 2021-04-14
Fix lint issues (0.1)

dplumb94:
Fixes issues noted by new clippy lints

### DIFF
--- a/contracts/pike/src/addresser.rs
+++ b/contracts/pike/src/addresser.rs
@@ -15,23 +15,23 @@
 /// Represents part of address that designates resource type
 #[derive(Debug)]
 pub enum Resource {
-    AGENT,
-    ORG,
+    Agent,
+    Org,
 }
 
 /// Convert resource part to byte value in hex
 pub fn resource_to_byte(part: Resource) -> String {
     match part {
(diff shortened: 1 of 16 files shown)


peterschwarz (dismissed):

RyanLassigBanks (approved):
Looks like there is still some lint. Will re approve when build passes

contracts/pike/src/addresser.rs
```diff
@@ -15,23 +15,23 @@
 /// Represents part of address that designates resource type
 #[derive(Debug)]
 pub enum Resource {
-    AGENT,
-    ORG,
+    Agent,
```
vaporos, line 18: Should this also be handled with #[allow(clippy::upper_case_acronyms)] to avoid the backward-incompatible change?

peterschwarz (approved):

RyanLassigBanks (approved):

contracts/pike/src/addresser.rs
dplumb94: I didn't think it would be appropriate since this addresser module is only used within the Pike contract. This file was removed in the dev branch and now pulls in the addressing module from the SDK.

PR #640, opened 2021-04-15
Product GDSN Validation in CLI

dplumb94:
This PR is based on #629. Remaining tasks include:

- Implementing update
- Fetching alternate ID to fill in the owner

To test: 

1. start up the `examples/splinter/docker-compose.yaml` network and create a Grid circuit between alpha and beta
2. Set up agents and organizations as follows:

```
grid keygen alpha-agent

export GRID_SERVICE_ID=<circuit ID>::gsAA

grid organization create crgl Cargill --alternate-ids gs1_company_prefix:0
grid role create crgl productowner --permissions schema::can-create-schema,product::can-create-product --active
grid agent update crgl $(cat ~/.grid/keys/alpha-agent.pub) --active --role productowner --role admin
```
3. submit a transaction to create a schema with the new `xsd_schema.yaml` file. If you are executing this from the gridd-alpha container you might need to copy it in. 
`$ grid schema create xsd-schema.yaml`
4. Create a product using some example XML data that conforms to the GridTradeItems XSD. This will go through validation and be created.
`$ grid product create -f yourproduct.yaml`
5. Create a product using some XML data that doesnt conform to the GridTradeItems XSD. The CLI should return an error telling you where the data was incorrect.


sdk/src/products/gdsn/validate.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    if let Err(errors) = xsd {
+        let mut xsd_errors: Vec<String> = vec![];
+
+        for err in &errors {
+            xsd_errors.push(err.message().to_string())
+        }
+
+        return Err(ProductGdsnError::Internal(InternalError::with_message(
+            format!("Error creating validation context: {:?}", xsd_errors),
```
vaporos, line 44: Consider using {} instead of {:?} as {:?} is giving the debug string. Same below.

dplumb94: Since this is a Vec<String> it looks like it cant be formatted with the default formatter

dplumb94: I updated it to pretty print {:#?} which makes the output a lot more readable

dnewh (approved):

vaporos (approved):

PR #644, opened 2021-04-19
Use "just" recipes for CI

rbuysse:


justfile
```diff
@@ -56,9 +56,59 @@ build-experimental:
...
+ci-test-ui: ci-build-ui-test-deps
+    #!/usr/bin/env sh
+    set -e
+    docker run --rm --env CI=true grid-ui:$ISOLATION_ID yarn test
+    docker run --rm --env CI=true product-sapling:$ISOLATION_ID yarn test
+
 clean:
     cargo clean
 
+integration-test:
```
vaporos, line 107: This should be ci-* an in-lined. The script seems overly complex compared to the above examples that don't call docker-compose down.

vaporos (changes requested):
The dockerfiles used here should also be modified to call just, to the extent practical.

PR #650, opened 2021-04-22
Add cross-cutting db features

dnewh:
This adds cross-cutting features to the Grid SDK for the locations, products, and schemas modules. These features control whether the db code for each module gets compiled.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -86,8 +86,10 @@ experimental = [
     "batch-store",
     "client",
     "client-reqwest",
-    "postgres",
+    "location-store-postgres",
+    "location-store-sqlite",
     "pike-rest-api",
+    "postgres",
     "product-gdsn",
     "purchase-order",
     "rest-api-resources",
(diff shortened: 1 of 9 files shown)


RyanLassigBanks (approved):

sdk/Cargo.toml
```diff
@@ -99,6 +101,8 @@ experimental = [
 client = []
 client-reqwest = ["client", "reqwest"]
 location = ["pike", "schema"]
+location-store-postgres = []
```
vaporos, line 108: This should depend on "postgresql" and similarly the line below should depend on "sqlite".

sdk/src/locations/mod.rs
```diff
@@ -13,8 +13,13 @@
 // limitations under the License.
 
 pub mod addressing;
+#[cfg(any(feature = "location-store-postgres", feature = "location-store-sqlite"))]
 pub mod store;
 
-#[cfg(feature = "diesel")]
+#[cfg(all(
+    feature = "diesel",
```
vaporos, line 20: diesel should be implied by the other two features; see comment on Cargo.toml. This comment applies to all other instances as well.

sdk/Cargo.toml
vaporos: same for product and schema

vaporos (changes requested):

RyanLassigBanks (approved):

vaporos (approved):

PR #655, opened 2021-04-29
Add last_updated field to rest api resources

RyanLassigBanks:
Adds database generated timestamp to agent, role, organization, product, location, and schema endpoints.

Example:

```
GET /agent/021bab739fce4133204c8c0bd939d48e4d32e5b29eb87ed871d2666fcd3f51bb3e
{
  "public_key": "021bab739fce4133204c8c0bd939d48e4d32e5b29eb87ed871d2666fcd3f51bb3e",
  "org_id": "crg",
  "active": true,
  "roles": [
    "admin"
  ],
  "metadata": {},
  "last_updated": 1619707637
}
```

### DIFF
--- a/sdk/src/migrations/diesel/sqlite/migrations/2021-04-23-151946_add_timestamp/down.sql
+++ b/sdk/src/migrations/diesel/sqlite/migrations/2021-04-23-151946_add_timestamp/down.sql
@@ -1,0 +1,39 @@
+-- Copyright 2021 Cargill Incorporated
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+-- -----------------------------------------------------------------------------
+
+DROP TRIGGER set_pike_agent_timestamp;
+DROP TRIGGER set_pike_organization_timestamp;
+DROP TRIGGER set_pike_role_timestamp;
+DROP TRIGGER set_grid_schema_timestamp;
+DROP TRIGGER set_product_timestamp;
+DROP TRIGGER set_location_timestamp;
+
+ALTER TABLE pike_agent
+DROP COLUMN last_updated;
+
+ALTER TABLE pike_organization
+DROP COLUMN last_updated;
+
+ALTER TABLE pike_role
+DROP COLUMN last_updated;
+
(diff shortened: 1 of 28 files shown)


sdk/src/migrations/diesel/sqlite/migrations/2021-04-23-151946_add_timestamp/down.sql
```diff
@@ -0,0 +1,88 @@
...
+-- -----------------------------------------------------------------------------
+
+DROP TABLE pike_agent;
+DROP TABLE pike_organization;
+DROP TABLE pike_role;
+DROP TABLE grid_schema;
+DROP TABLE product;
+DROP TABLE location;
+
+CREATE TABLE pike_agent (
```
dnewh, line 23: Could these be ALTER TABLE commands rather than DROPping and CREATEing?

dnewh (commented):
tested and it works, just had the comment regarding the migrations

sdk/src/migrations/diesel/sqlite/migrations/2021-04-23-151946_add_timestamp/down.sql
RyanLassigBanks: You can't use ALTER TABLE with DEFAULT CURRENT_TIMESTAMP in sqlite. You get this error 
```
Schema Error: Error: SQLITE_ERROR: Cannot add a column with non-constant default
```
Unfortunately, the only way around it is to recreate the table.

dnewh (dismissed):

RyanLassigBanks:
> DROP/CREATE approach here will delete data in existing installs.

We also take this approach for the pike 2 migration, and we can't add the date created field to the sqlite tables using ALTER TABLE (see my response to @dnewh). Do you know of an alternative solution?

dplumb94 (approved):

vaporos:
> We also take this approach for the pike 2 migration...

That was an explicit decision to drop backward compatibility for Pike 1 and the upgrade path will need to be documented/tested in the release notes.

sdk/src/migrations/diesel/sqlite/migrations/2021-04-23-151946_add_timestamp/down.sql
vaporos: IIRC, its pretty common to create a table then copy the data to the new table, rename, etc.

Perhaps remove this default and use a trigger anyway, allowing this column to be NULL for existing data.

peterschwarz: This isn't used for dates, but the fundamental principle applies: https://github.com/Cargill/splinter/blob/main/libsplinter/src/migrations/diesel/sqlite/migrations/2021-02-09-042200_update-identity-enum/up.sql

dplumb94 (changes requested):
`Add date_created field to gridd endpoints` commit message should be updated since that field has a different name now

dnewh (approved):

dplumb94 (approved):

PR #656, opened 2021-05-03
Rename `fetch_` operations to `get_`

dnewh:
This renames the `fetch_` operations to `get_` operations. This brings
the Grid data store operation in line with the naming conventions used
in Splinter. This is part of stabilizing the postgres and sqlite features.


dplumb94 (commented):
I get that this is how they are named in Splinter but I don't think it is ideal. Fetch and List correspond nicely to the URI functions whereas Get seems to be overloaded with the HTTP method name.  My 2 cents.

dnewh:
> I get that this is how they are named in Splinter but I don't think it is ideal. Fetch and List correspond nicely to the URI functions whereas Get seems to be overloaded with the HTTP method name. My 2 cents.

We've discussed this a couple times now. I think the decision was to rename to `get_*`

agunde406 (approved):

RyanLassigBanks (changes requested):
I was under the impression that this change was only to pertain to the store functions, which I also disagree with changing, but have begrudgingly accepted.

RyanLassigBanks (approved):

PR #658, opened 2021-05-03
Make all operations atomic

dnewh:
This makes all operations atomic. Previously, this was done in a sort of
piecemeal fashion. This just cleans this up and makes sure all
operations are wrapped in the generic diesel transaction. This is
available for all Diesel backends that implement Diesel::Connection.
This depends on PR #656.

PR #689, opened 2021-05-13
Have fetch_role use transaction

RyanLassigBanks:
Have postgres implementation of fetch_role use transaction instead of
the transaction build, and have the sqlite implementation use
transaction instead of immediate_transaction. This is to make the
implementations consistent with the other store operations.


dnewh (changes requested):
This is already covered in PR #658

PR #692, opened 2021-05-14
List and show agents in the CLI

KevinAnsleyJohnson:
Includes man-pages. List includes formatting human and csv types, along
with the ability to list roles on their own lines. The Show command
gives additional information, and requires a public key as the
identifier. (Agent ID is synonymous with public key.)

### DIFF
--- a/cli/src/actions/agents.rs
+++ b/cli/src/actions/agents.rs
@@ -15,14 +15,30 @@
  * -----------------------------------------------------------------------------
  */
 
+use crate::actions::ListSlice;
 use crate::error::CliError;
 use crate::http::submit_batches;
 use crate::transaction::pike_batch_builder;
+use reqwest::Client;
+use serde::Deserialize;
+use std::cmp;
+use std::collections::HashMap;
+
 use grid_sdk::{
     pike::addressing::PIKE_NAMESPACE,
     protocol::pike::payload::{Action, CreateAgentAction, PikePayloadBuilder, UpdateAgentAction},
     protos::IntoProto,
 };
+
+#[derive(Debug, Deserialize)]
+pub struct AgentSlice {
+    pub public_key: String,
+    pub org_id: String,
+    pub active: bool,
+    pub roles: Vec<String>,
+    pub service_id: Option<String>,
+    pub metadata: HashMap<String, String>,
+}
 
 pub fn do_create_agent(
     url: &str,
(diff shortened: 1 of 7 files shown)


RyanLassigBanks:
about for agent subcommand should be updated to reference list and how
https://github.com/hyperledger/grid/blob/5541e3518a583d269b90eeba520bc984babc1610/cli/src/main.rs#L156

KevinAnsleyJohnson:
> about for agent subcommand should be updated to reference list and show
> https://github.com/hyperledger/grid/blob/5541e3518a583d269b90eeba520bc984babc1610/cli/src/main.rs#L156

good catch! however, this also is the case for many other commands as well. I may make a separate PR and fix them all, but for now I will update the current one to reflect the correct fix.

cli/src/actions/agents.rs
```diff
@@ -15,15 +15,31 @@
...
     protos::IntoProto,
 };
 
+#[derive(Debug, Deserialize)]
+pub struct AgentSlice {
+    pub public_key: String,
+    pub org_id: String,
+    pub active: bool,
+    pub roles: Vec<String>,
+    pub service_id: String,
```
RyanLassigBanks, line 39: `service_id` should be `Option<String>` since the service_id` will only be present if running in splinter mode. If grid is connect to a sawtooth network the cli will fail to deserialize this payload as the api will not return a `service_id`

RyanLassigBanks (changes requested):

cli/src/actions/agents.rs
KevinAnsleyJohnson: Thanks! Also, since service_id is only for splinter, I think it makes sense to remove it from the `show` option as well, otherwise there would be inconsistency. PR is now updated reflecting the Options<String> change, as well as removing the line that prints the Service ID in `show`.

dplumb94 (approved):
Tested, LGTM.

sample output for other reviewers:

```
MacBook-Pro:grid dplumb$ grid agent list
PUBLIC_KEY                                                         ORG_ID ACTIVE ROLES                                                  
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   productowner, admin, role1, role2, role3, role4, role5 
MacBook-Pro:grid dplumb$ grid agent list --line-per-role
PUBLIC_KEY                                                         ORG_ID ACTIVE ROLE         
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   productowner 
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   admin        
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   role1        
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   role2        
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   role3        
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   role4        
02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778 crgl   true   role5        
MacBook-Pro:grid dplumb$ grid agent list --format=csv
"PUBLIC_KEY","ORG_ID","ACTIVE","ROLES"
"02583654747f593243a7d8635397cc8bbdd25540d26ae1e4318dd99141a03aa778","crgl","true","productowner, admin, role1, role2, role3, role4, role5"
MacBook-Pro:grid dplumb$ grid agent list --format=asdf
error: 'asdf' isn't a valid value for '--format <format>'
        [possible values: csv, human]


USAGE:
    grid agent list --format <format>
```

RyanLassigBanks (approved):
Tested locally

PR #697, opened 2021-05-18
Refactor pike protocol payload

RyanLassigBanks:
PikePayload protocol object now has a single action object instead of
multiple payload objects. The Action enum has been updated to hold a
payload object that corresponds to each enum type.

### DIFF
--- a/sdk/src/protocol/pike/payload.rs
+++ b/sdk/src/protocol/pike/payload.rs
@@ -1586,11 +1586,16 @@ impl DeleteRoleActionBuilder {
 #[derive(Debug, Clone, PartialEq)]
 pub struct PikePayload {
     action: Action,
+    timestamp: u64,
 }
 
 impl PikePayload {
     pub fn action(&self) -> &Action {
         &self.action
+    }
+
+    pub fn timestamp(&self) -> u64 {
+        self.timestamp
     }
 }
 
@@ -1639,13 +1644,16 @@ impl FromProto<protos::pike_payload::PikePayload> for PikePayload {
             }
         };
 
-        Ok(Self { action })
+        let timestamp = payload.get_timestamp();
+
+        Ok(Self { action, timestamp })
     }
 }
 
 impl FromNative<PikePayload> for protos::pike_payload::PikePayload {
     fn from_native(native: PikePayload) -> Result<Self, ProtoConversionError> {
         let mut proto = protos::pike_payload::PikePayload::new();
+        proto.set_timestamp(native.timestamp());
 
         match native.action() {
             Action::CreateAgent(payload) => {
@@ -1748,6 +1756,7 @@ impl std::fmt::Display for PikePayloadBuildError {
 #[derive(Default, Clone)]
 pub struct PikePayloadBuilder {
     pub action: Option<Action>,
+    pub timestamp: Option<u64>,
 }
 
 impl PikePayloadBuilder {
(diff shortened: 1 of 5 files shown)


sdk/src/protocol/pike/payload.rs
```diff
@@ -26,71 +26,17 @@ use crate::protos::{
 /// Native implementation for PikePayload_Action
 #[derive(Debug, Clone, PartialEq)]
 pub enum Action {
```
agunde406, line 28: Action could just be PikePayload since PikePayload only has the action as a field.

agunde406 (changes requested):

dplumb94 (approved):

sdk/src/protocol/pike/payload.rs
RyanLassigBanks: I think this change makes sense, but that would make the builder kind of pointless, so I would suggest removing that as well. Also this would make the structure of the pike payload substantially different from the payload structs for the other protocols. If people are ok with then, I'm cool with making this change. Thoughts @vaporos @dplumb94 @dnewh ?

dplumb94: Hmm it looks like the other payload structs have a timestamp field. I think it would actually make more sense to add that timestamp field to this transaction to match the others.

agunde406: If that gets added, then I'm fine with it staying like this.

ui/protos/pike_payload.proto
```diff
@@ -34,6 +34,8 @@ message PikePayload {
 
   CreateOrganizationAction create_organization = 4;
   UpdateOrganizationAction update_organization = 5;
+
+  uint64 timestamp = 6;
```
agunde406, line 38: This proto file does not match the other proto file changed above, should this just be updated in the future when the whole protobuf is updated?

sdk/src/protocol/pike/payload.rs
```diff
@@ -1586,12 +1586,17 @@ impl DeleteRoleActionBuilder {
 #[derive(Debug, Clone, PartialEq)]
 pub struct PikePayload {
     action: Action,
+    timestamp: u64,
```
peterschwarz, line 1589: Should the field on protocol object be a `Duration`?  The u64/seconds would be at the protobuf level, and it would give some strongly-typed meaning here.

Granted, it should also match the other instances of timestamps. If they are all currently u64's, then perhaps a story could be opened to cover changing all of them to the appropriate field type.

ui/protos/pike_payload.proto
RyanLassigBanks: Reverted changes to this file

sdk/src/protocol/pike/payload.rs
```diff
@@ -1760,12 +1769,21 @@ impl PikePayloadBuilder {
...
+        self
+    }
+
     pub fn build(self) -> Result<PikePayload, PikePayloadBuildError> {
         let action = self.action.ok_or_else(|| {
             PikePayloadBuildError::MissingField("'action' field is required".to_string())
         })?;
 
-        Ok(PikePayload { action })
+        let timestamp = self.timestamp.ok_or_else(|| {
```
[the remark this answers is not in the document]
RyanLassigBanks: This is meant to be consistent with the implementations [ProductPayloadBuilder](https://github.com/hyperledger/grid/blob/f0842a908da7f10ecc2540a4afc18e1090954f72/sdk/src/protocol/product/payload.rs#L178) and [LocationPayloadBuilder](https://github.com/hyperledger/grid/blob/f0842a908da7f10ecc2540a4afc18e1090954f72/sdk/src/protocol/location/payload.rs#L213)

sdk/src/protocol/pike/payload.rs
RyanLassigBanks: All the other instances of timestamps in the protocol module are `u64`.

agunde406 (approved):

sdk/src/protocol/pike/payload.rs
peterschwarz: Opened #710 to cover this detail.

dplumb94 (approved):

PR #698, opened 2021-05-19
Update signer to match splinter's pattern

KevinAnsleyJohnson:
Perform signer logic before passing to handlers.  Signer implementation
follows Splinter's implementation pattern. This also introduces a new
overload, CYLINDER_PATH, if GRID_DAEMON_KEY is not set, and if --key
option is not included.  create_cylinder_jwt_auth method is also
included for future implementation, but for now is decorated with an
`allow dead_code` attribute.


cli/src/actions/locations.rs
```diff
@@ -12,13 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{
```
peterschwarz, line 15: Some of these use statements have been re-ordered unnecessarily.

cli/src/actions/agents.rs
```diff
@@ -66,7 +66,7 @@ pub fn do_create_agent(
 
 pub fn do_update_agent(
     url: &str,
-    key: Option<String>,
+    signer: Box<dyn Signer>,
```
peterschwarz, line 69: Do functions need to own the signer?  If not, this could be typed `&dyn Signer`.

cli/src/actions/locations.rs
KevinAnsleyJohnson: That's the rust linter's preference :)

cli/src/actions/agents.rs
KevinAnsleyJohnson: I'm not sure whether the functions need to own the signer or not - it was done this way to align with Splinter's cli.

cli/src/signing.rs
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    Ok(private_key)
+}
+
+pub fn load_signer(key_name: Option<String>) -> Result<Box<dyn Signer>, CliError> {
+    Ok(Secp256k1Context::new().new_signer(load_private_key(key_name)?))
+}
+
+#[allow(dead_code)]
```
vaporos, line 86: Please exclude jwt support entirely for now or guard jwt support with an experimental feature "cylinder-jwt".

cli/Cargo.toml
```diff
@@ -32,7 +32,7 @@ path = "src/main.rs"
 assert_cmd = "1.0"
 clap = "2"
 diesel = { version = "1.0", features = ["postgres"], optional = true }
-cylinder = "0.2"
+cylinder = { version = "0.2.2", features = ["jwt", "key-load"] }
```
vaporos, line 35: See jwt comment below. If kept, pull in jwt support via a dependency on an experimental feature.

cli/src/actions/agents.rs
```diff
@@ -19,16 +19,16 @@ use crate::actions::ListSlice;
...
+use cylinder::Signer;
 use grid_sdk::{
     pike::addressing::PIKE_NAMESPACE,
     protocol::pike::payload::{Action, CreateAgentAction, PikePayloadBuilder, UpdateAgentAction},
     protos::IntoProto,
 };
+use reqwest::Client;
+use serde::Deserialize;
+use std::cmp;
+use std::collections::HashMap;
```
vaporos, line 31: Since this re-orders the use statements, it would be good to conform to grouping them in the order of: std, external crates, crate, module with empty between the groupings (to prevent rust fmt from reordering). Unfortunately we don't have a lint for this.

cli/src/actions/locations.rs
vaporos: Best practice (for this project anyway) is to let the reviewer resolve their own comments as a double-check.

The re-ordering occurred because of the removal of the newline. See the previous comment on the right grouping of use statements.

cli/src/transaction.rs
```diff
@@ -142,20 +144,25 @@ impl BatchBuilder {
         }
         output_addresses.append(&mut outputs.to_vec());
 
-        let private_key = key::load_signing_key(self.key_name.clone())?;
-        let context = signing::create_context("secp256k1")?;
-        let public_key = context.get_public_key(&private_key)?.as_hex();
-        let factory = signing::CryptoFactory::new(&*context);
-        let signer = factory.new_signer(&private_key);
+        // let private_key = self.signer.()?; //key::load_signing_key(self.key_name.clone())?;
```
vaporos, line 147: Commented out code isn't allowed in the project.

vaporos (changes requested):

cli/src/actions/agents.rs
vaporos: I don't recall why I didn't do &dyn Signer in Splinter, but IIRC I did attempt that and ended up with Box<dyn Signer> because I couldn't easily resolve some compiler issue without modifying cylinder.

cli/src/actions/locations.rs
KevinAnsleyJohnson: Ah! thanks for letting me know (about both things). Will fix!

cli/Cargo.toml
KevinAnsleyJohnson: I think it is better to remove in this case, because there truly is no implementation.

cli/src/actions/agents.rs
KevinAnsleyJohnson: Ok! Sounds good. will fix!

cli/src/transaction.rs
KevinAnsleyJohnson: completed.

cli/Cargo.toml
KevinAnsleyJohnson: completed.

vaporos (approved):

peterschwarz (approved):

PR #712, opened 2021-05-25
Add Schema namespace to location txn inputs

dplumb94:
This was causing location transactions to fail when the smart contract
attempted to read the location schema.

Signed-off-by: Darian Plumb <dplumb@bitwise.io>


peterschwarz (approved):

peterschwarz:
Any way to add a test to catch a regression?

agunde406 (approved):

dplumb94:
@peterschwarz  We do have CLI tests, but I believe the problem here is in Scabbard since it is not checking the inputs properly for Splinter deployments. I'll check if any of the integration tests would cover this.

PR #719, opened 2021-05-27
Remove sqlite and postgres from rest api feature

vaporos:
This removes sqlite and postgres from the sdk's rest-api-actix-web-3
feature, which will make it possible to conditionally use the sqlite or
postgresql when using the rest api.

This commit makes BatchList pub, because without pub, it would be
necessary to put a conditional of the underlying implementations being
enabled; it seems cleaner to simply make it pub since it does not seem
intentional to prevent creation of BatchList (which itself is pub).

Features are added to griddle for sqlite/postgres to pull in the
associated sdk features.

### DIFF
--- a/griddle/src/main.rs
+++ b/griddle/src/main.rs
@@ -18,12 +18,16 @@ extern crate log;
 mod error;
 
 use std::env;
+#[cfg(any(feature = "database-postgres", feature = "database-sqlite"))]
 use std::str::FromStr;
 use std::sync::Arc;
 
 use clap::{App, Arg};
+#[cfg(feature = "diesel")]
 use diesel::r2d2::{ConnectionManager, Pool};
 use flexi_logger::{DeferredNow, LogSpecBuilder, Logger};
+#[cfg(any(feature = "database-postgres", feature = "database-sqlite"))]
+use grid_sdk::store::ConnectionUri;
 use grid_sdk::{
     batch_processor::{
         submitter::{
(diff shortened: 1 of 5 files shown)


sdk/Cargo.toml
```diff
@@ -129,18 +129,16 @@ schema-store-postgres = ["postgres"]
 schema-store-sqlite = ["sqlite"]
 track-and-trace = ["base64"]
 batch-processor = ["batch-store", "backend", "log", "uuid"]
-batch-store = []
+batch-store = ["chrono"]
```
peterschwarz, line 132: This change seems unrelated

vaporos: It is related, because with the change chrono is no longer pulled into the batch-store non-diesel code via sqlite/postgres implementation if they are disabled.

dnewh (approved):

peterschwarz (dismissed):

griddle/src/main.rs
```diff
@@ -47,11 +50,20 @@ fn log_format(
     write!(w, "{}", record.args(),)
 }
 
+#[cfg(not(any(feature = "database-postgres", feature = "database-sqlite")))]
+fn griddle_store_state(_db_url: &str) -> Result<StoreState, Error> {
+    Err(Error::from_message(
+        "no database feature was enabled durinig compilation",
```
dnewh, line 56: s/durinig/during

dnewh (changes requested):

griddle/src/main.rs
vaporos: fixed

dnewh (approved):

peterschwarz (approved):

PR #729, opened 2021-06-03
Update help with env var defaults

KevinAnsleyJohnson:
`Key` option help information updated to include GRID_DAEMON_PATH and
CYLINDER_PATH defaults. `service_id` option help was updated to include
GRID_SERVICE_ID environment variable. `url` option help was updated to
include GRID_SERVICE_ENDPOINT default.

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -97,7 +97,18 @@ purchase-order = ["pike"]
 rest-api = ["database"]
 sawtooth-support = ["database", "event", "rest-api"]
 schema = ["pike"]
-splinter-support = ["database", "event", "rest-api", "reqwest", "scabbard", "sabre-sdk", "splinter", "transact/contract-archive", "serde_json", "serde"]
+splinter-support = [
+  "database",
+  "event",
+  "reqwest",
+  "rest-api",
+  "sabre-sdk",
+  "scabbard",
+  "serde_json",
+  "serde",
+  "splinter",
+  "transact/contract-archive",
+]
 test-postgres = []
 track-and-trace = []
 integration = []
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -71,7 +71,8 @@ fn run() -> Result<(), DaemonError> {
         (@arg database_url: --("database-url") +takes_value
          "specifies the database URL to connect to.")
         (@arg bind: -b --bind +takes_value "connection endpoint for rest API")
-        (@arg admin_key_dir: --("admin-key-dir") +takes_value "directory containing the Scabbard admin key files"));
+        (@arg admin_key_dir: --("admin-key-dir") +takes_value
+            "directory containing the Scabbard admin key files"));
 
     #[cfg(feature = "integration")]
     {
@@ -115,7 +116,8 @@ fn run() -> Result<(), DaemonError> {
             }
             #[cfg(not(feature = "sawtooth-support"))]
             Err(DaemonError::with_message(&format!(
-                "A Sawtooth connection endpoint ({}) was provided but Sawtooth support is not enabled for this binary.",
+                "A Sawtooth connection endpoint ({}) was provided but Sawtooth support is not \
+                enabled for this binary.",
                 config.endpoint().url()
             )))
         }
@@ -127,7 +129,8 @@ fn run() -> Result<(), DaemonError> {
             }
             #[cfg(not(feature = "splinter-support"))]
             Err(DaemonError::with_message(&format!(
-                "A Splinter connection endpoint ({}) was provided but Splinter support is not enabled for this binary.",
+                "A Splinter connection endpoint ({}) was provided but Splinter support is not \
+                enabled for this binary.",
                 config.endpoint().url()
             )))
         }


cli/src/main.rs
```diff
@@ -304,16 +316,6 @@ fn run() -> Result<(), CliError> {
                     .subcommand(
                         SubCommand::with_name("list")
                             .about("List all agents for a given service")
-                            .arg(
```
KevinAnsleyJohnson, line 307: Noticed that this implementation was duplicating already existing code above. so it was redundant that therefore removed.

cli/src/main.rs
```diff
@@ -1170,10 +1244,18 @@ fn run() -> Result<(), CliError> {
                             .takes_value(true)
                             .help("Workflow status of this Purchase Order version"),
                     )
-                    .arg(Arg::with_name("draft").long("draft").help(
-                        "Specify this Purchase Order version is not a draft. \
-                        By default, a newly created version is a draft.",
-                    ))
+                    .arg(
```
KevinAnsleyJohnson, line 1247: Man page specified that there would be both a `draft` - which is by default, and a `not-draft`. Updated this implementation to match the man-page, because draft is a flag and is set to true by default, there would be no way of setting it to false. Also draft and not-draft is the convention used for other components. This method is not yet implemented so it should have no impact on the application.

cli/src/main.rs
vaporos: Can you break this out to a separate commit?

cli/src/main.rs
vaporos: Can you break this out into a separate commit?

cli/src/main.rs
```diff
@@ -222,7 +226,11 @@ fn run() -> Result<(), CliError> {
                                 .long("key")
                                 .short("k")
                                 .takes_value(true)
-                                .help("Base name for private signing key file")
+                                .help(
+                                    "Base name for private signing key file. (either a file path or the name of a \
+                                    .priv file in $HOME/.grid/keys). This location can also be set via the \
+                                    GRID_DAEMON_PATH variable, and if not present may be set with CYLINDER_PATH \
```
vaporos, line 232: Was this intended to be GRID_DAEMON_KEY instead of GRID_DAEMON_PATH?

vaporos (changes requested):
In the past, we have resisted extremely long usage lines (as this is very common in Linux command usage statements, trying to target more "git commit -h" than "git commit --help".

The trade-off is essentially between concise quick-reference and verbose completeness, and we've previously favored concise with the assumption the man page covers completeness. 

ew have at times however, used a "DETAILS:" section to expand upon how the command works if its truly necessary.

For example, the most complex command across projects is "splinter circuit propose" which looks something like:

```
splinter-circuit-propose
Propose that a new circuit is created

USAGE:
    splinter circuit propose [FLAGS] [OPTIONS] --node <node>... --node-file <node_file> --service <service>... --template <template>

FLAGS:
    -n, --dry-run    Print circuit definition without submitting the proposal
    -h, --help       Prints help information
    -q, --quiet      Do not display output
    -V, --version    Prints version information
    -v               Log verbosely

OPTIONS:
        --comments <comments>                           Add human-readable comments to the proposal
        --compat <compat_version>
            Enforce that the proposed circuit is compatible with a specific version [possible values: 0.4, 0.6]

        --display-name <display_name>                   Add human-readable name for the circuit
    -k, --key <private-key-file>                        Path to private key file
        --management <management_type>                  Management type for the circuit
        --metadata <application_metadata>...            Application metadata of the proposal
        --metadata-encoding <metadata_encoding>
            Set encoding of application metadata (default: string) [possible values: json, string]

        --node <node>...
            Node that is part of a circuit (<node_id>::<endpoint1>,<endpoint2>)

        --node-file <node_file>                         File system path or HTTP(S) URL to nodes file
        --service <service>...                          Service ID and allowed nodes (<service-id>::<allowed_nodes>)
        --service-arg <service_argument>...             Pass arguments to a service (<service_id>::<key>=<value>)
        --service-peer-group <service_peer_group>...    List of peer services (comma-separated list)
        --service-type <service_type>...                Service type (<service_id>::<service_type>)
        --template <template>                           Template name to be applied to circuit
        --template-arg <template_arg>...                Arguments for the template argument (<key>=<value>)
    -U, --url <url>                                     URL of Splinter Daemon

DETAILS:
    One or more nodes must be specified using the --node and/or --node-file arguments. These
    arguments can be used on their own or together, but at least one of them is required.

    The --node-file argument must be a valid YAML file. A valid YAML file will be a list of nodes,
    where each node has an 'identity' or 'node_id' field, as well as an 'endpoints' field. Example:
        ---
        - identity: 'node-1'
          endpoints:
            - tcps://node-1-endpoint:8044
        - node_id: 'node-2'
          endpoints:
            - tcps://node-2-endpoint:8045

    For the --service-arg, --service-peer-group, and --service-type options, service IDs can be
    wildcarded with '*' to match multiple services. For example, '--service-type *::scabbard' match
    all services, and '--service-type sc*::scabbard' will match all services with IDs that start
    with 'sc'.

    With '--metadata-encoding string' (the default), the --metadata option takes a single string
    value and the --metadata option can be used only once. With '--metadata-encoding json', the
    --metadata option takes a key/value pair in the format '<key>=<value>', where '<value>' is a
    simple string, a JSON array, or a JSON object; the --metadata option can be used multiple times
    with JSON encoding.
```

cli/src/main.rs
KevinAnsleyJohnson: Absolutely - I was wondering that myself.

cli/src/main.rs
KevinAnsleyJohnson: yes - Will remove from this PR

KevinAnsleyJohnson: This is now removed.

cli/src/main.rs
KevinAnsleyJohnson: Ok - This is now removed, and will be in a separate PR

cli/src/main.rs
KevinAnsleyJohnson: Yes - now fixed! Thanks!

PR #734, opened 2021-06-04
Remove redundant service-id option at subcommand

KevinAnsleyJohnson:
Noticed that this implementation was duplicating already existing code
at the parent subcommand, so it was redundant and therefore removed.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -305,16 +305,6 @@ fn run() -> Result<(), CliError> {
                         SubCommand::with_name("list")
                             .about("List all agents for a given service")
                             .arg(
-                                Arg::with_name("service_id")
-                                    .long("service-id")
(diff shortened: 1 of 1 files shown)


dplumb94 (approved):

vaporos (changes requested):
This should be specified in both places. It should be ok to do:

grid agent --service-id=a list

or

grid agent list --service-id=a

KevinAnsleyJohnson:
> This should be specified in both places. It should be ok to do:
> 
> grid agent --service-id=a list
> 
> or
> 
> grid agent list --service-id=a

I believe it already works like this. No other subcommand needs to redefine. I will check locally and let you know.

KevinAnsleyJohnson:
Closing due to service-id being able to be defined on either side of the argument

Issue #330, opened 2020-09-14
grid CLI: environment variables missing from usage


chenette:
The grid CLI usage (--help output) should mention the available environment variables:

- `GRID_DAEMON_ENDPOINT` (instead of specifying `--url`)
- `GRID_DAEMON_KEY` (instead of `-k`)
- `GRID_SERVICE_ID` (instead of `--service_id`)

Best practice is to briefly mention the environment variable in the option's description.  If more information is needed, it can be included in the "after_help" section of the usage, but this is optional and should also be very brief.  (The man page will have full information on using the environment variable.)

For example, see the Splinter CLI, such as `splinter cert generate --help`:

> -d, --cert-dir <cert_dir>          
>             Path to the directory certificates are created in. Defaults to /etc/splinter/certs/. This location can also
>             be changed with the SPLINTER_CERT_DIR environment variable. This directory must exist.

PR #735, opened 2021-06-04
Update help with env var defaults

KevinAnsleyJohnson:
`Key` option help information updated to include GRID_DAEMON_KEY and
CYLINDER_PATH defaults. `service_id` option help was updated to include
GRID_SERVICE_ID environment variable. URL option help was updated to
include GRID_SERVICE_ENDPOINT default.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -73,6 +73,30 @@ const GRID_SERVICE_ID: &str = "GRID_SERVICE_ID";
 
 const SYSTEM_KEY_PATH: &str = "/etc/grid/keys";
 
+#[cfg(any(
+    feature = "location",
+    feature = "pike",
+    feature = "product",
+    feature = "schema",
+    feature = "purchase-order"
+))]
+const AFTER_HELP_WITHOUT_KEY: &str = r"ENV:
+    GRID_DAEMON_ENDPOINT   Specifies a default value for --url
+    GRID_SERVICE_ID        Specifies a default value for --service-id";
+
+#[cfg(any(
+    feature = "location",
+    feature = "pike",
+    feature = "product",
+    feature = "schema",
+    feature = "purchase-order"
+))]
+const AFTER_HELP_WITH_KEY: &str = r"ENV:
+    CYLINDER_PATH          Path to search for private signing keys
+    GRID_DAEMON_ENDPOINT   Specifies a default value for --url
+    GRID_DAEMON_KEY        Specifies a default value for -k, --key
+    GRID_SERVICE_ID        Specifies a default value for --service-id";
+
 // log format for cli that will only show the log message
 pub fn log_format(
     w: &mut dyn std::io::Write,
@@ -229,7 +253,8 @@ fn run() -> Result<(), CliError> {
                                 .long("wait")
                                 .takes_value(true)
                                 .help("How long to wait for transaction to be committed")
-                        ),
+                        )
+                        .after_help(AFTER_HELP_WITH_KEY)
                 )
                 .subcommand(
                     SubCommand::with_name("update")
@@ -289,7 +314,8 @@ fn run() -> Result<(), CliError> {
                                 .long("wait")
                                 .takes_value(true)
                                 .help("How long to wait for transaction to be committed")
-                        ),
+                        )
+                        .after_help(AFTER_HELP_WITH_KEY),
                     )
                     .subcommand(
                         SubCommand::with_name("show")
@@ -300,6 +326,7 @@ fn run() -> Result<(), CliError> {
                                     .required(true)
                                     .help("Public Key and unique identifier for agents"),
                             )
+                            .after_help(AFTER_HELP_WITHOUT_KEY),
                     )
                     .subcommand(
                         SubCommand::with_name("list")
@@ -329,6 +356,7 @@ fn run() -> Result<(), CliError> {
                                     .help("Displays agent information for each role on it's own \
                                         line. Useful when filtering by role.")
                             )
+                            .after_help(AFTER_HELP_WITHOUT_KEY),
                     ),
         )
         .subcommand(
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -1170,10 +1227,6 @@ fn run() -> Result<(), CliError> {
                             .takes_value(true)
                             .help("Workflow status of this Purchase Order version"),
                     )
-                    .arg(Arg::with_name("draft").long("draft").help(
```
dplumb94, line 1173: was this meant to be in a different commit?

dplumb94 (commented):
This approach does match previous implementations, but im wondering if we should use the https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.env method instead. We also wouldnt have to manually extract the env vars if we used it.

cli/src/main.rs
KevinAnsleyJohnson: My Mistake! yes - it's now updated.

vaporos (changes requested):
This is PR is a dup of #729.

See the comment in #729 about general approach we've been taking with the CLI -- preferring reference readability over completeness.

KevinAnsleyJohnson:
> This is PR is a dup of #729.
> 
> See the comment in #729 about general approach we've been taking with the CLI -- preferring reference readability over completeness.

Looking at your comment on 729. I made updates to reduce the help length, but the story I have assigned to me asked for the help to contain the environment variables (https://github.com/hyperledger/grid/issues/330). Should we update the issue, or is there a different compromise you have in mind?

vaporos:
> This approach does match previous implementations, but im wondering if we should use the https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.env method instead. We also wouldnt have to manually extract the env vars if we used it.

This is possibly interesting, but there are a few possible negatives. The primary negative is that environment variables and tests don't mix, so if we go this route, we may be deciding to not write tests against clap code. When we have the integration tests, it might actually be desirable to call main() (essentially) of the CLI code from within the tests to get full coverage end-to-end within the tests. (Even now, we would need to implement a way to turn off environment variable handling, but that's relatively easy.) The most obvious reason is you can't have multiple tests in the same process setting the same environment variables or you will get unpredictable behavior. It is also harder to log the source of the message in output, like splinterd does, and that can be very useful if users get confused about what the actual configuration is and how it got loaded. There might also be cases where it simply doesn't work because of how things are evaluated.

Personally, my preferred pattern for this is something similar to what we have in splinterd, where we take each source individually, (cli, env, config, default, etc.) and turn it into an intermediate "partial" config representation and then merge those into the final configuration. It is very clean but has the downside of needing to update code in several files when an option is added or changed.

vaporos:
A middle-ground here is to keep the options simple but add an ENV section with after_help():

```
ENV:
    GRID_SERVICE_ID        Specify a default value for --service-id
    GRID_SERVICE_ENDPOINT  Specify a default value for --url
```

This probably shouldn't be added when there are SUBCOMMANDS though -- because putting it after SUBCOMMANDS makes it harder to read, so this would only be for "leaf" commands.

cli/src/main.rs
```diff
@@ -229,6 +229,13 @@ fn run() -> Result<(), CliError> {
                                 .long("wait")
                                 .takes_value(true)
                                 .help("How long to wait for transaction to be committed")
+                        )
+                        .after_help(
+                            "ENV:\n\n\
+                                CYLINDER_PATH          Path to search for private signing keys\n\
+                                GRID_DAEMON_ENDPOINT   Specifies a default value for --url\n\
+                                GRID_DEAMON_KEY        Specifies a default value for -k, --key\n\
```
dnewh, line 237: s/DEAMON/DAEMON

This also occurs in several other places in this PR

dnewh (changes requested):

cli/src/main.rs
KevinAnsleyJohnson: Thanks! It's been updated now!

KevinAnsleyJohnson:
> I'm seeing indentation like this:
> 
> ```
> grid-agent-list
> List all agents for a given service
> 
> USAGE:
>     grid agent list [FLAGS] [OPTIONS]
> 
> FLAGS:
>     -h, --help             Prints help information
>         --line-per-role    Displays agent information for each role on it's own line. Useful when filtering by role.
>     -q, --quiet            Do not display output
>     -V, --version          Prints version information
>     -v                     Log verbosely
> 
> OPTIONS:
>     -F, --format <format>            Output format [default: human]  [possible values: human, csv]
>         --service-id <service_id>    The ID of the service the payload should be sent to; required if running on
>                                      Splinter. Format <circuit-id>::<service-id>
> 
> ENV:
> 
> GRID_DAEMON_ENDPOINT   Specifies a default value for --url
> GRID_SERVICE_ID        Specifies a default value for --service-id
> ```
> 
> expected:
> 
> ```
> grid-agent-list
> List all agents for a given service
> 
> USAGE:
>     grid agent list [FLAGS] [OPTIONS]
> 
> FLAGS:
>     -h, --help             Prints help information
>         --line-per-role    Displays agent information for each role on it's own line. Useful when filtering by role.
>     -q, --quiet            Do not display output
>     -V, --version          Prints version information
>     -v                     Log verbosely
> 
> OPTIONS:
>     -F, --format <format>            Output format [default: human]  [possible values: human, csv]
>         --service-id <service_id>    The ID of the service the payload should be sent to; required if running on
>                                      Splinter. Format <circuit-id>::<service-id>
> 
> ENV:
>     GRID_DAEMON_ENDPOINT   Specifies a default value for --url
>     GRID_SERVICE_ID        Specifies a default value for --service-id
> ```

Totally Agree. I was following splinter, but this makes sense. Updated!

dnewh (approved):

vaporos (approved):

PR #736, opened 2021-06-04
Update man pages for agent commands

KevinAnsleyJohnson:

### DIFF
--- a/cli/man/grid-agent-create.1.md
+++ b/cli/man/grid-agent-create.1.md
@@ -14,7 +14,13 @@ NAME
 SYNOPSIS
 ========
 
-**grid agent create** \[**FLAGS**\] \[**OPTIONS**\] ORG_ID PUBLIC_KEY --active --inactive
+**grid agent create** \[**FLAGS**\] \[**OPTIONS**\] <ORG_ID> <PUBLIC_KEY> <{**--active**|**--inactive**}>
+
+DESCRIPTION
+===========
+
+This command creates a new agent. ORG_ID and PUBLIC_KEY arguments are required, 
+as well as the --active or --inactive flag.
 
 ARGS
 ====
@@ -40,21 +46,28 @@ FLAGS
 `-v`
 : Log verbosely
 
+`--active`
+: sets agent as active. conflicts with `--inactive`
+
+`--inactive`
+: sets agent as inactive. conflicts with `--active`
+
 OPTIONS
 =======
 
 `-k`, `--key`
-: Base name for private signing key file
+: base name or path to a private signing key file
 
 `--metadata`
 : Key-value pairs (format: <key>=<value>) in a comma-separated list
 
 `--role`
-: Roles assigned to the agent
+: Roles assigned to the agent. Multiple roles can be assigned in a 
+  comma-separated list.
 
 `--service-id`
 : The ID of the service the payload should be sent to; required if running on
-Splinter. Format <circuit-id>::<service-id>
+  Splinter. Format <circuit-id>::<service-id>.
 
 `--url`
 : URL for the REST API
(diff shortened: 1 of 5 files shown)


cli/man/grid-agent-create.1.md
```diff
@@ -40,34 +46,49 @@ FLAGS
...
+  .priv file in $HOME/.grid/keys). This location can also be set via the 
+  GRID_DAEMON_KEY variable, and if not present may be set with CYLINDER_PATH
+  variable.
 
 `--metadata`
 : Key-value pairs (format: <key>=<value>) in a comma-separated list
 
 `--role`
-: Roles assigned to the agent
+: Roles assigned to the agent. May specify more than once.
```
dnewh, line 68: This also accepts a comma-separated list. Perhaps this should be included in this description too

KevinAnsleyJohnson: that's true - qq: should it be named "roles" instead of "role"?

cli/man/grid-agent.1.md
```diff
@@ -38,9 +38,15 @@ FLAGS
 ENVIRONMENT VARIABLES
 =====================
 
+Many subcommands accept the following environment variables:
+
+**`CYLINDER_PATH`**
+: Path used to sign transactions if `k` or `--key`
```
vaporos, line 44: Perhaps: "Colon-separated path used to search for the key which will be used to sign transactions."

I'm not sure how GRID_DAEMON_KEY is implemented, but if it is a straight-up replacement for "-k/--key", then there are multiple ways that it might interact with CYLINDER_PATH. If a "/" is present, it is interpreted as either a relative path or absolute path to the key file itself. If it does not include a "/" then it is the "base name" of the key, and all the directories listed in CYLINDER_PATH will be searched (in order) for a file called KEY.priv.

cli/man/grid-agent-create.1.md
```diff
@@ -40,34 +46,49 @@ FLAGS
...
+
+`--inactive`
+: sets agent as inactive. conflicts with `--active`
+
 OPTIONS
 =======
 
 `-k`, `--key`
-: Base name for private signing key file
+: Base name for private signing key file. (either a file path or the name of a
```
vaporos, line 59: "Base name" here is not technically correct, since if it has a "/" then it is interpreted as the full path to the file. In other words, the key option should:

--key=mykey # search all directories in CYLINDER_PATH for mykey.priv
--key=./mykey.priv # use the file ./mykey.priv
--key=path/mykey.priv # use the file ./path/mykey.priv
etc.

The second sentence should start with an E and the period should be within the parens, or the period preceding the parens should be omitted (depending on intent).

cli/man/grid-agent-create.1.md
vaporos: If it can be specified using a delimiter, then it could be changed to the plural form. We've kept aliases for the non-plural form in other places, but hidden them and not documented them.

vaporos (changes requested):

cli/man/grid-agent.1.md
KevinAnsleyJohnson: Updated!

cli/man/grid-agent-create.1.md
KevinAnsleyJohnson: Updated!

cli/man/grid-agent-create.1.md
KevinAnsleyJohnson: description is now updated to include comma-separated list. pluralization of a cli command should be a separate commit, if we want to do it.

dnewh (approved):

vaporos (approved):

PR #741, opened 2021-06-04
Use in-memory db for rest api tests

dnewh:
This updates the rest api tests to use an in-memory sqlite db. This
offers a couple benefits. One, this "resets" the database between tests
so that the db does not need to be cleared manually. Two, this does not
create any database files that would need to be manually cleaned up
after the tests.


peterschwarz (commented):
Make sure it is added to any dockerignore files

KevinAnsleyJohnson (dismissed):
LGTM

vaporos (changes requested):
The test should be modified to create this in a temporary directory and then clean it up after.

vaporos (approved):

daemon/src/rest_api/mod.rs
```diff
@@ -767,25 +742,12 @@ mod test {
...
-                srv.url(&format!("/agent?service_id={}", TEST_SERVICE_ID)),
-            )
-            .send()
-            .await
-            .unwrap();
-        assert!(response.status().is_success());
-        let body: AgentListSlice =
-            serde_json::from_slice(&*response.body().await.unwrap()).unwrap();
-        assert!(body.data.is_empty());
+        // let pool = create_connection_pool_and_migrate();
```
dplumb94, line 745: remove commented code

dplumb94 (changes requested):

daemon/src/rest_api/mod.rs
dnewh: fixed

dplumb94 (approved):
Worked locally, LGTM

vaporos (approved):

PR #745, opened 2021-06-07
Add accessor methods for data structures

dnewh:
This adds public accessor methods with doc strings for each of the
public data structures in the Product module. This also makes the
corresponding fields private.

### DIFF
--- a/sdk/src/product/store/mod.rs
+++ b/sdk/src/product/store/mod.rs
@@ -20,48 +20,486 @@ use crate::paging::Paging;
 
 #[cfg(feature = "diesel")]
 pub use self::diesel::DieselProductStore;
-pub use error::ProductStoreError;
+pub use error::{ProductBuilderError, ProductStoreError};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Product {
-    pub product_id: String,
-    pub product_address: String,
-    pub product_namespace: String,
-    pub owner: String,
-    pub start_commit_num: i64,
-    pub end_commit_num: i64,
-    pub service_id: Option<String>,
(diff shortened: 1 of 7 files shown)


sdk/src/product/store/mod.rs
```diff
@@ -20,49 +20,487 @@ use crate::paging::Paging;
...
+    pub fn product_id(&self) -> &str {
+        &self.product_id
+    }
+
+    /// Returns the product_address for the product
+    pub fn product_address(&self) -> &str {
+        &self.product_address
+    }
+
+    /// Returns the product_namespace for the product
```
shannynalayna, line 49: These comments don't seem necessary, I think adding comments like this will get to be a lot with all of the stores.

sdk/src/product/store/mod.rs
```diff
@@ -20,49 +20,487 @@ use crate::paging::Paging;
...
+        self
+    }
+
+    /// Sets the product namespace for this product
+    pub fn with_product_namespace(mut self, product_namespace: String) -> Self {
+        self.product_namespace = product_namespace;
+        self
+    }
+
+    /// Returns the owner of the product
```
shannynalayna, line 118: Copy-pasta here with the comment

sdk/src/product/store/mod.rs
dnewh: In Splinter, these all have doc comments. I'm fine with omitting them here because I don't think these really add much but we should discuss whether or not to break from the splinter pattern for things like this.

peterschwarz (approved):

dplumb94 (approved):

peterschwarz (approved):

PR #750, opened 2021-06-09
Fix app auth handler

dplumb94:
A recent refactor of the app auth handler introduced some regressions. This PR fixes them.

### DIFF
--- a/daemon/src/splinter/app_auth_handler/sabre.rs
+++ b/daemon/src/splinter/app_auth_handler/sabre.rs
@@ -90,6 +90,10 @@ pub fn setup_grid(
     #[cfg(feature = "pike")]
     make_pike_txns(&mut txns, &version, &signer)?;
 
+    // Make schema transactions
+    #[cfg(feature = "schema")]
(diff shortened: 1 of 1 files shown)


dnewh (approved):

shannynalayna (approved):

daemon/src/splinter/app_auth_handler/sabre.rs
```diff
@@ -90,6 +90,10 @@ pub fn setup_grid(
     #[cfg(feature = "pike")]
     make_pike_txns(&mut txns, &version, &signer)?;
 
+    // Make schema transactions
+    #[cfg(feature = "schema")]
```
peterschwarz, line 94: Isn’t this only an issue when the contracts are executed?

dplumb94: The Schema namespace has to exist before the Product and Location CreateNamespaceRegistryPermissionAction transactions for the Schema namespace are executed. Since the schema transactions were added after product and location, this was causing an error.

PR #752, opened 2021-06-09
Pike smart contract stabilization updates

shannynalayna:
This PR makes various updates to the Pike smart contract required for stabilization. Mainly:
- Move `PikeState` to its own module
- Replaces instances of for-loops with the iter's `find` method, where possible, in the `get_*` state methods
- Updates the copyright headers to accurately reflect the year range the files were created/updated
- Several other small updates to comments and consts, as requested


contracts/pike/src/state.rs
```diff
@@ -0,0 +1,461 @@
...
+                        err,
+                    )))
+                }
+            },
+            None => AgentList::new(),
+        };
+        // remove old agent if it exists and sort the agents by public key
+        let agents = agent_list.get_agents().to_vec();
+        let mut index = None;
+        for (i, agent) in agents.iter().enumerate() {
```
KevinAnsleyJohnson, line 222: could you use the iterator's position method here? https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position

Also on lines 296 and 371. Adding comments at those places as well.

contracts/pike/src/state.rs
```diff
@@ -0,0 +1,461 @@
...
+                        err,
+                    )))
+                }
+            },
+            None => OrganizationList::new(),
+        };
+        // remove old organization if it exists and sort the organizations by org ID
+        let organizations = organization_list.get_organizations().to_vec();
+        let mut index = None;
+        for (i, organization) in organizations.iter().enumerate() {
```
KevinAnsleyJohnson, line 296: I believe you could also use the iterator's position method here. (https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position)

contracts/pike/src/state.rs
```diff
@@ -0,0 +1,461 @@
...
+                        "Cannot deseralize alternate ID index entry list: {:?}",
+                        err,
+                    )));
+                }
+            },
+            None => vec![],
+        };
+
+        let mut index = None;
+        for (i, e) in entries.iter().enumerate() {
```
KevinAnsleyJohnson, line 371: Iterator's position method here too.

contracts/pike/src/state.rs
vaporos: This code is being moved, and we should keep any enhancements separate from the reorganization.

contracts/pike/src/state.rs
vaporos: This code is being moved, and we should keep any enhancements separate from the reorganization.

contracts/pike/src/state.rs
vaporos: This code is being moved, and we should keep any enhancements separate from the reorganization.

vaporos (approved):

agunde406 (approved):

PR #754, opened 2021-06-09
Check all required schema for product / location

KevinAnsleyJohnson:
Instead of short-circuiting at any CliError in yaml_to_property_values
functions, iterate through the entire payload and return a more
comprehensive list of errors. Because we're coalescing the CliError
types, I chose UserError to encapsulate the differing types of CliErrors
that may occur. Also Introduced InternalError to the cli,
already present in grid_sdk.

### DIFF
--- a/cli/src/actions/locations.rs
+++ b/cli/src/actions/locations.rs
@@ -194,118 +194,139 @@ fn yaml_to_property_values(
     definitions: Vec<PropertyDefinition>,
 ) -> Result<Vec<PropertyValue>, CliError> {
     let mut property_values = Vec::new();
+    let mut property_error_messages = Vec::new();
 
     for def in definitions {
-        let value = if let Some(value) = properties.get(&def.name) {
-            value
-        } else if !def.required {
-            continue;
-        } else {
-            return Err(CliError::PayloadError(format!(
-                "Field {} not found",
-                def.name
-            )));
-        };
-
-        match def.data_type {
-            DataType::Bytes => {
-                let mut f = File::open(&serde_yaml::from_value::<String>(value.clone())?)?;
-                let mut buffer = Vec::new();
-                f.read_to_end(&mut buffer)?;
-
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name)
-                    .with_data_type(def.data_type.into())
-                    .with_bytes_value(buffer)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-
-                property_values.push(property_value);
+        match get_property_value_from_property_definition(properties, def) {
+            Ok(Some(property_value)) => property_values.push(property_value),
+            Ok(None) => (),
+            Err(CliError::YamlProcessingError(mut messages)) => {
+                property_error_messages.append(&mut messages);
             }
-            DataType::Boolean => {
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_boolean_value(serde_yaml::from_value(value.clone())?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
+            Err(CliError::PayloadError(message)) => {
+                property_error_messages.push(message);
             }
-            DataType::Number => {
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_number_value(serde_yaml::from_value(value.clone())?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
-            }
-            DataType::String => {
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_string_value(serde_yaml::from_value(value.clone())?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
-            }
-            DataType::Enum => {
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_enum_value(serde_yaml::from_value(value.clone())?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
-            }
-            DataType::Struct => {
-                let properties: HashMap<String, serde_yaml::Value> =
-                    serde_yaml::from_value(value.clone())?;
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_struct_values(yaml_to_property_values(
-                        &properties,
-                        def.struct_properties,
-                    )?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
-            }
-            DataType::LatLong => {
-                let lat_long = serde_yaml::from_value::<String>(value.clone())?
-                    .split(',')
-                    .map(|x| {
-                        x.parse::<i64>()
-                            .map_err(|err| CliError::PayloadError(format!("{}", err)))
-                    })
-                    .collect::<Result<Vec<i64>, CliError>>()?;
-
-                if lat_long.len() != 2 {
-                    return Err(CliError::PayloadError(format!(
-                        "{:?} is not a valid latitude longitude",
-                        lat_long
-                    )));
-                }
-
-                let lat_long = LatLongBuilder::new()
-                    .with_lat_long(lat_long[0], lat_long[1])
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name)
-                    .with_data_type(def.data_type.into())
-                    .with_lat_long_value(lat_long)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-
-                property_values.push(property_value);
+            Err(err) => {
+                return Err(err);
             }
         }
     }
 
-    Ok(property_values)
+    if !property_error_messages.is_empty() {
+        Err(CliError::YamlProcessingError(property_error_messages))
+    } else {
+        Ok(property_values)
+    }
+}
+
+fn get_property_value_from_property_definition(
+    properties: &HashMap<String, serde_yaml::Value>,
+    def: PropertyDefinition,
+) -> Result<Option<PropertyValue>, CliError> {
+    let value = if let Some(value) = properties.get(&def.name) {
+        value
+    } else if !def.required {
+        return Ok(None);
+    } else {
+        return Err(CliError::PayloadError(format!(
+            "Field {} not found",
+            def.name
+        )));
+    };
+
+    match def.data_type {
+        DataType::Bytes => {
+            let mut f = File::open(&serde_yaml::from_value::<String>(value.clone())?)?;
+            let mut buffer = Vec::new();
+            f.read_to_end(&mut buffer)?;
+
+            let property_value = PropertyValueBuilder::new()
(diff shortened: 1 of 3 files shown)


cli/src/actions/locations.rs
```diff
@@ -233,118 +233,146 @@ fn yaml_to_property_values(
...
-                    .with_name(def.name)
-                    .with_data_type(def.data_type.into())
-                    .with_bytes_value(buffer)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
+    if errors.len() > 0 {
+        let list_of_errors = errors
+            .iter()
+            .map(|e| match e {
+                CliError::UserError(str) => Some(str.to_string()),
```
KevinAnsleyJohnson, line 250: Wondering - should I exhaust the match cases? I believe UserError and PayloadError capture all possible errors, but including IoError and InvalidYamlError because we do file opening and yaml parsing on line 285.

cli/src/actions/locations.rs
```diff
@@ -233,118 +233,146 @@ fn yaml_to_property_values(
...
+                CliError::PayloadError(str) => Some(str.to_string()),
+                CliError::InvalidYamlError(str) => Some(str.to_string()),
+                CliError::IoError(str) => Some(str.to_string()),
+                _ => None,
+            })
+            .filter_map(|e| e)
+            .collect::<Vec<String>>()
+            .join("\n    - ");
+        Err(CliError::InvalidYamlError(format!(
```
KevinAnsleyJohnson, line 259: I chose InvalidYamlError as an overarching error. I could alternatively send multiple errors back, but then would need more handling.

cli/src/actions/products.rs
```diff
@@ -571,119 +571,148 @@ fn yaml_to_property_values(
...
+    let value = if let Some(value) = properties.get(&def.name) {
+        value
+    } else if !def.required {
+        return Ok(None);
+    } else {
+        #[cfg(feature = "product-gdsn")]
+        if def.name == GDSN_3_1_PROPERTY_NAME {
+            return Ok(None);
+        }
+        return Err(CliError::UserError(format!("Field {} not found", def.name)));
```
KevinAnsleyJohnson, line 619: The original function listed this as a UserError, but would this ultimately be a Payload Error?

cli/src/actions/locations.rs
dplumb94: Hmm yeah this is tricky. Ultimately I think the top level error should be UserError. The thing the user really needs to know is the errors that come from checking the submitted product definition against the schema. Things like IO errors or payload errors (coming from serializing the protobuf) are internal and not really actionable by the user, so it might not make sense to include them in this list. It might make more sense to just short circuit and return one of those errors at that point.

cli/src/actions/locations.rs
```diff
@@ -201,7 +201,15 @@ pub fn create_location_payloads_from_file(
             Namespace::Gs1 => "gs1_location",
         };
         let schema = get_schema(url, namespace, service_id)?;
-        payloads.push(yml.into_payload(schema.properties)?);
+        payloads.push(
+            yml.into_payload(schema.properties)
+                .map_err(|err| match err {
+                    CliError::EmptyError => {
```
KevinAnsleyJohnson, line 207: Notice the new type: EmptyError - This is in response to our discussion of how many times we have errors that are never propagated to the CLI. Not sold on this approach, but offering it more as a RFC.

cli/src/actions/locations.rs
```diff
@@ -233,118 +249,140 @@ fn yaml_to_property_values(
...
             }
-            schemas::DataType::Enum => {
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name.clone())
-                    .with_data_type(def.data_type.into())
-                    .with_enum_value(serde_yaml::from_value(value.clone())?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
+            Err(CliError::EmptyError) => {
```
KevinAnsleyJohnson, line 262: Because of the recursive nature of the function and its helper, we don't want to report the same error at multiple nested levels. We just want to note that an error occurred, and have confidence that the causes of EmptyError (Payload or User Errors) have been reported via the `error!` macro. an emptyError occurring at any level of nesting will eventually be handled with proper verbiage higher up on the call stack, as found on lines 204 and 233.

cli/src/actions/locations.rs
vaporos: We should consider removing CliError entirely in favor of InternalError taken from libsplinter.

Here is some high-level info: https://www.splinter.dev/community/error_handling_best_practices.html

We could start this by adding InternalError to the CliError enum:

```
enum CliError {
...
    InternalError(InternalError),
...
}
```

And over time, reducing that enum as we collapse things into InternalError, and potentially removing CliError completely at some point.

Other errors could be used more locally, where we need a function to return an enum so we can differentiate between multiple errors; but the end result would always be to eventually convert to an InternalError and start propagating that up the stack.

It is important to note the meaning of "InternalError" -- "this error has no meaning to the calling function, other than that an error occurred". It does not mean "not caused by the user".

vaporos (changes requested):

cli/src/actions/locations.rs
KevinAnsleyJohnson: Brilliant!! I totally agree with this approach. Thanks Shawn!

cli/src/error.rs
```diff
@@ -32,6 +32,7 @@ pub enum CliError {
     GridProtoError(protos::ProtoConversionError),
     SabreProtoError(sabre_sdk::protos::ProtoConversionError),
     DaemonError(String),
+    InternalError(String),
```
vaporos, line 35: This variant should be InternalError(InternalError) where the InternalError struct is the one from the Grid SDK.

KevinAnsleyJohnson: Thank Shawn! It's now updated.

cli/src/actions/products.rs
```diff
@@ -564,118 +581,143 @@ fn yaml_to_property_values(
...
-                    )?)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-                property_values.push(property_value);
+        }
+    }
+
+    if are_errors_present {
+        Err(CliError::InternalError(InternalError::with_message(
+            "Nested Error".to_string(),
```
KevinAnsleyJohnson, line 605: because this function is recursive, an InternalError firing N deep into the yaml tree would report N times, hence we want to ignore this with an internal error. At the top level we capture the internal error and respond that there was an error with the specific file.

KevinAnsleyJohnson:
## Testing this branch:

bash into gridd-alpha with proper GRID_SERVICE_ID and GRID_DAEMON_KEY set.
Run the setup scripts:
```
echo '  - name: gs1_location
    description: GS1 location schema
    owner: myorg
    properties:
      - name: locationDescription
        data_type: String
        description: ""
        required: true
      - name: locationType
        data_type: Enum
        description: ""
        enum_options: [
          ShipTo, BillTo, ShipFrom, PaidBy, OrderFrom, Recall, OrgEntity, RemitTo
        ]
        required: true
      - name: addressLine1
        data_type: String
        description: ""
        required: true
      - name: addressLine2
        data_type: String
        description: ""
        required: false
      - name: city
        data_type: String
        description: ""
        required: true
      - name: stateOrRegion
        data_type: String
        description: ""
        required: true
      - name: postalCode
        data_type: String
        description: ""
        required: true
      - name: country
        data_type: String
        description: ""
        required: true
      - name: latLongValue
        data_type: lat_long
        description: ""
        required: true
      - name: contactName
        data_type: String
        description: ""
        required: true
      - name: contactEmail
        data_type: String
        description: ""
        required: true
      - name: contactPhone
        data_type: String
        description: ""
        required: true
      - name: createDate
        data_type: Number
        number_exponent: 0
        description: ""
        required: true
      - name: inactivationDate
        data_type: Number
        number_exponent: 0
        description: ""
        required: false
      - name: parentLocation
        data_type: String
        description: ""
        required: false
      - name: industrySector
        data_type: Enum
        description: ""
        enum_options: [General, CPG, Healthcare, Foodservice]
        required: false
      - name: role
        data_type: Enum
        description: ""
        enum_options: [
          Manufacturer, SolutionsProvider, Undefined, Distributor,
          Provider, Supplier, 3rdParty, Warhouse, IndependentOperator, Operator]
        required: false
      - name: informationProviderGLN
        data_type: Number
        number_exponent: 0
        description: ""
        required: false' >> location_schema.yaml;

echo ' - namespace: "GS1"
   location_id: "0013600000011"
   owner: "myorg"
   properties:
     locationDescription: "test location"
     locationType: 0
     addressLine1: "251 Test St."
     city: "Minneapolis"
     stateOrRegion: "MN"
     postalCode: "55401"
     country: "United States"
     latLongValue: "449778,-932650"
     contactName: "John Tester"
     contactEmail: "tester@email.com"
     contactPhone: "6555555555"
     createDate: 1602706399'>> good_location.yaml;

echo ' - namespace: "GS1"
   location_id: "0013600000011"
   owner: "myorg"
   properties:
     wronglocationDescription: "test location"
     locationType: 0
     wrongaddressLine1: "251 Test St."
     wrongcity: "Minneapolis"
     wrongstateOrRegion: "MN"
     postalCode: "55401"
     country: "United States"
     latLongValue: "wrong449778,wrong-932650"
     contactName: "John Tester"
     contactEmail: "tester@email.com"
     contactPhone: "6555555555"
     createDate: 1602706399'>> bad_location.yaml;
```

then run the execution:
```
    grid schema create location_schema.yaml;
    grid location create -f good_location.yaml;
    grid location create -f bad_location.yaml;
```

good_location.yaml should post correctly, but
bad_location.yaml should fail, and show multiple errors

cli/src/actions/locations.rs
```diff
@@ -162,7 +163,15 @@ pub fn create_location_payloads_from_file(
             Namespace::Gs1 => "gs1_location".to_string(),
         };
         let schema = client.get_schema(namespace, service_id)?;
-        payloads.push(yml.into_payload(schema.properties)?);
+        payloads.push(
+            yml.into_payload(schema.properties)
+                .map_err(|err| match err {
+                    CliError::InternalError(_) => {
+                        CliError::UserError(format!("Errors occurred while processing: {}", path))
```
vaporos, line 170: To avoid throwing away the internal error message, this could be - ("Errors occurred while processing '{}': {}", path, e)

cli/src/actions/products.rs
```diff
@@ -501,118 +518,143 @@ fn yaml_to_property_values(
...
-                let property_value = PropertyValueBuilder::new()
-                    .with_name(def.name)
-                    .with_data_type(def.data_type.into())
-                    .with_bytes_value(buffer)
-                    .build()
-                    .map_err(|err| CliError::PayloadError(format!("{}", err)))?;
-
-                property_values.push(property_value);
+            Err(CliError::InternalError(_)) => {
+                are_errors_present = true;
```
vaporos, line 532: This should probably be logged with error!() as well.

KevinAnsleyJohnson: I think an example would be helpful.

example:

- namespace: "GS1"
   location_id: "0013600000011"
   owner: "myorg"
   properties:
     locationDescription: "test location"
     locationType: 0
     addressLine1: "251 Test St."
     city: "Minneapolis"
     stateOrRegion: "" // error: stateOrRegion required
     postalCode: "55401"
     country: "United States"
     latLongValue: "449778," // error: cannot parse latLong
     contactName: "John Tester"
     contactEmail: "tester@email.com"
     contactPhone: "6555555555"
     createDate: 1602706399
  
this payload would show errors related to:
 - stateOrRegion required
 - cannot parse latLong

but, because it's nested under properties, we just want to notify the parent that an error occurred.

otherwise, it would show 3 errors:
 - stateOrRegion required
 - cannot parse latLong
 - Nested error

hence, the nested error is a necessary propogation tool. Another way to look at this is to remove the error state entirely, and instead send a valid flag to it's parent.

cli/src/actions/locations.rs
KevinAnsleyJohnson: Internal Error is only used as a propagation tool within the current context. If that's causing confusion, we could try to pass whether the state is valid or not.

shannynalayna:
Based on changing priorities and that this PR is out-of-date, it will be closed. Feel free to re-open when the merge conflicts are fixed, after rebasing to main, and existing comments are addressed. This PR may also need to be backported to 0-3 if re-opened on main.

PR #772, opened 2021-06-18
Update location ops

dnewh:
This fixes a bug in the operation to update a location in the Grid DB. This bug was causing the `end_commit_num` to not be updated. This also removes the unused `update_location` operation. Updates to locations are handled by the `add_location` operation. This is in line with the other Grid DB modules.


KevinAnsleyJohnson:
Should we rename add_location to add_update_location or upsert_location? something to be more descriptive of the fact that add_location also updates?

dnewh:
> Should we rename add_location to add_update_location or upsert_location? something to be more descriptive of the fact that add_location also updates?

When we had this discussion internally, we decided against that due to the slowly-changing dimensions. when updating a record in Grid, the existing db entry is updated but a new entry is also created with the `end_commit_num` set to MAX. This is the record that is considered current.

vaporos (approved):

dplumb94 (approved):

PR #792, opened 2021-06-29
Generate man pages for cli

KevinAnsleyJohnson:
Uses pandoc to convert markdown files from cli/man to manfiles in
cli/packaging/man. This follows the same pattern found in the Splinter
project's cli.


vaporos (changes requested):
The commit message should contain information about the origin of this approach/code (copied from Splinter, etc., etc.).

agunde406 (approved):

dplumb94 (approved):

PR #797, opened 2021-07-02
Fix various small issues in Pike CLI

dplumb94:


cli/src/main.rs
```diff
@@ -320,28 +320,18 @@ fn run() -> Result<(), CliError> {
...
                                     .required(true)
-                                    .help("Public Key and unique identifier for agents"),
+                                    .help("Public key and unique identifier for agents"),
                             )
                             .after_help(AFTER_HELP_WITHOUT_KEY),
                     )
                     .subcommand(
                         SubCommand::with_name("list")
                             .about("List all agents for a given service")
-                            .arg(
```
KevinAnsleyJohnson, line 335: This is one that I tried removing before and Shawn said it should stay - basically allows the --service-id to come before or after the arguments... Although this is not standard to the rest of the cli

KevinAnsleyJohnson (commented):
Awesome!!!

Just a quick note on one area.

cli/src/main.rs
dplumb94: Fixed by making the the service-id and url args global one level up from here.

cli/src/main.rs
```diff
@@ -185,6 +185,7 @@ fn run() -> Result<(), CliError> {
                     Arg::with_name("service_id")
                         .long("service-id")
                         .takes_value(true)
+                        .global(true)
```
KevinAnsleyJohnson, line 188: very cool!

cli/src/main.rs
```diff
@@ -528,7 +528,7 @@ fn run() -> Result<(), CliError> {
                                 .takes_value(true)
                                 .multiple(true)
                                 .use_delimiter(true)
-                                .help("List of organizations allowed use of the role"),
+                                .help("List of org IDs of organizations allowed use of the role"),
```
peterschwarz, line 537: Is "org" redundant?

vaporos: I agree with this - everything else LGTM

peterschwarz (approved):

shannynalayna (approved):

PR #807, opened 2021-07-12
Update smart contracts to version 2

rbuysse:
This change is required to make them compatible with the new pike addressing.


dplumb94 (changes requested):
Not sure why it isnt using the consts but we need to upgrade the Location batch builder too. Line 67 of cli/src/transaction.rs.

dnewh (approved):

dplumb94 (approved):

PR #834, opened 2021-07-19
Reqwest Client SDK Move wip - do not merge

KevinAnsleyJohnson:


sdk/src/client/mod.rs
```diff
@@ -31,117 +37,69 @@ pub trait Client {
...
-    ///
-    /// * `id` - The uuid of the `PurchaseOrder` containing the `PurchaseOrderVersion` to be retrieved
-    /// * `version_id` - The version id of the `PurchaseOrderVersion` to be retrieved
-    #[cfg(feature = "purchase-order")]
-    fn get_purchase_order_version(
-        &self,
-        id: String,
-        version_id: String,
-    ) -> Result<Option<PurchaseOrderVersion>, InternalError>;
+pub trait ClientFactory {
```
shannynalayna, line 49: For clients, I think it makes more sense to have individual `builder` structs for each client implemented. So, within the client module, we should probably have a `builder` module (final organization/location of the builder modules depends on how we separate the clients) where the `_ClientBuilder` would be defined

shannynalayna:
Ok, so I like where the separate clients angle is going but I think we may want to re-think how the clients are separated. I don't think it makes sense to make different clients for each type of resource defined (org, agent, location, etc.) but instead perhaps by the smart contracts that oversee those objects since we're separating things by smart contract throughout grid. 

So, I think in this case the client organization would look more like `sdk/client/pike` and in the mod.rs file you would define the `PikeClient` trait. Then, in `sdk/client/pike/reqwest` would be the Reqwest-backed implementation of that client. The pike client would contain methods to access orgs, agents, roles, so pike-related objects. And then the same so on and so forth for each smart contract. 

In terms of the resources that are used to deserialize the rest responses and what is available in our public API, these should be separate. For the top-level client trait definitions, these methods should return a non-(de)serializable struct. Then, in the reqwest modules for each client, we can define client-specific structs that are able to deserialize (i.e., have serde's deserialize trait) the rest responses. These structs would then be able to be converted (pretty simply, the main difference here is the publicly available resource structs shouldn't have serde traits) to the top-level resource struct once the response has been deserialized.

KevinAnsleyJohnson (commented):
> Ok, so I like where the separate clients angle is going but I think we may want to re-think how the clients are separated. I don't think it makes sense to make different clients for each type of resource defined (org, agent, location, etc.) but instead perhaps by the smart contracts that oversee those objects since we're separating things by smart contract throughout grid.
> 
> So, I think in this case the client organization would look more like `sdk/client/pike` and in the mod.rs file you would define the `PikeClient` trait. Then, in `sdk/client/pike/reqwest` would be the Reqwest-backed implementation of that client. The pike client would contain methods to access orgs, agents, roles, so pike-related objects. And then the same so on and so forth for each smart contract.
> 
> In terms of the resources that are used to deserialize the rest responses and what is available in our public API, these should be separate. For the top-level client trait definitions, these methods should return a non-(de)serializable struct. Then, in the reqwest modules for each client, we can define client-specific structs that are able to deserialize (i.e., have serde's deserialize trait) the rest responses. These structs would then be able to be converted (pretty simply, the main difference here is the publicly available resource structs shouldn't have serde traits) to the top-level resource struct once the response has been deserialized.

I like that Idea - I'd like to discuss this option with the team.

KevinAnsleyJohnson:
> > Ok, so I like where the separate clients angle is going but I think we may want to re-think how the clients are separated. I don't think it makes sense to make different clients for each type of resource defined (org, agent, location, etc.) but instead perhaps by the smart contracts that oversee those objects since we're separating things by smart contract throughout grid.
> > So, I think in this case the client organization would look more like `sdk/client/pike` and in the mod.rs file you would define the `PikeClient` trait. Then, in `sdk/client/pike/reqwest` would be the Reqwest-backed implementation of that client. The pike client would contain methods to access orgs, agents, roles, so pike-related objects. And then the same so on and so forth for each smart contract.
> > In terms of the resources that are used to deserialize the rest responses and what is available in our public API, these should be separate. For the top-level client trait definitions, these methods should return a non-(de)serializable struct. Then, in the reqwest modules for each client, we can define client-specific structs that are able to deserialize (i.e., have serde's deserialize trait) the rest responses. These structs would then be able to be converted (pretty simply, the main difference here is the publicly available resource structs shouldn't have serde traits) to the top-level resource struct once the response has been deserialized.
> 
> I like that Idea - I'd like to discuss this option with the team.

On second look, I'm curious about the fact the API layer has already abstracted out the details of which smart-contract is being used. would we be reintroducing a detail that the client should not necessarily be aware of?

PR #886, opened 2021-07-30
Fix 1.54.0 lint

shannynalayna:
This PR includes several commits to fix new clippy lint errors from the Rust 1.54.0 release.

### DIFF
--- a/sdk/src/track_and_trace/store/diesel/operations/list_associated_agents.rs
+++ b/sdk/src/track_and_trace/store/diesel/operations/list_associated_agents.rs
@@ -65,9 +65,7 @@ impl<'a> TrackAndTraceStoreListAssociatedAgentsOperation
                 TrackAndTraceStoreError::NotFoundError(
                     "Could not get all records from storage".to_string(),
                 )
-            })?
-            .into_iter()
-            .collect();
+            })?;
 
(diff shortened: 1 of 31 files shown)


peterschwarz (approved):

sdk/src/track_and_trace/store/diesel/operations/list_associated_agents.rs
```diff
@@ -107,9 +105,7 @@ impl<'a> TrackAndTraceStoreListAssociatedAgentsOperation
                 TrackAndTraceStoreError::NotFoundError(
                     "Could not get all records from storage".to_string(),
                 )
-            })?
-            .into_iter()
-            .collect();
+            })?;
 
         Ok(models.into_iter().map(AssociatedAgent::from).collect())
```
peterschwarz, line 110: Is this into_iter still needed?

shannynalayna: It didn't come up with any clippy warnings, but I can try removing the into_iter here - will update based on results

shannynalayna: Looks like it is, as the list comes out of the query as a vec

vaporos (approved):

dnewh (approved):

PR #888, opened 2021-07-30
Remove unused docker compose files

shannynalayna:
This change removes unused docker compose files used to set up Grid with
Sawtooth and Splinter backends.

These docker compose files have been replaced by compose files in
specific `examples` modules.


dnewh (approved):

shannynalayna:
Seems like the docker-compose thought to be unused is used at a point in CI, further investigation into the use of the compose files is necessary before they are able to be removed.

PR #893, opened 2021-08-03
Avoid duplicate events

peterschwarz:
This change adds an EventProcessors collection, which ensures that only a single event processor on a given circuit/service connection is only created once.

It also adds a commit to serialize the processing of events as they go to the database. This will work around a current issue with transaction creation.


dplumb94 (dismissed):
tested, lgtm

dplumb94 (dismissed):

PR #897, opened 2021-08-06
Refactor SDE transaction handling

dnewh:
This PR updates the SDE to properly run database operations in a transaction. To facilitate this, two new `StoreFactory` traits were introduced, `TransactionalStoreFactory` and `InContextStoreFactory`, that expose this functionality in the daemon. These traits allow for the creation of a `StoreFactory` that can be used to initialize a transaction, get various Grid stores, and then run DB operations all within a transaction on a single Connection. Previously, there was no guarantee that operations were run within the same transaction as each operation got a Connection from a ConnectionPool individually.

### DIFF
--- a/daemon/src/database/mod.rs
+++ b/daemon/src/database/mod.rs
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Cargill Incorporated
+ * Copyright 2019-2021 Cargill Incorporated
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,15 +50,6 @@ impl<C: diesel::Connection> ConnectionPool<C> {
                 }
             })?,
         })
-    }
-
-    pub fn get(&self) -> Result<PooledConnection<ConnectionManager<C>>, DatabaseError> {
-        self.pool
-            .get()
-            .map_err(|err| DatabaseError::ConnectionError {
-                context: "Failed to get Connection from connection pool".to_string(),
-                source: Box::new(err),
-            })
     }
 }
 
--- a/daemon/src/event/db_handler.rs
+++ b/daemon/src/event/db_handler.rs
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Cargill Incorporated
+ * Copyright 2019-2021 Cargill Incorporated
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,8 +15,6 @@
  * -----------------------------------------------------------------------------
  */
 
-use diesel::Connection;
-use grid_sdk::commits::{store::CommitEvent as DbCommitEvent, CommitStore, DieselCommitStore};
 #[cfg(any(feature = "location", feature = "product"))]
 use grid_sdk::protocol::schema::state::PropertyValue;
 #[cfg(any(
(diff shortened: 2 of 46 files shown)


daemon/src/database/mod.rs
```diff
@@ -52,7 +52,7 @@ impl<C: diesel::Connection> ConnectionPool<C> {
         })
     }
 
-    pub fn get(&self) -> Result<PooledConnection<ConnectionManager<C>>, DatabaseError> {
+    pub fn _get(&self) -> Result<PooledConnection<ConnectionManager<C>>, DatabaseError> {
```
shannynalayna, line 55: Could this function just be removed?

daemon/src/splinter/app_auth_handler/mod.rs
```diff
@@ -68,6 +68,7 @@ pub fn run(
     let node_id = get_node_id(splinterd_url.clone())?;
 
     let ws_handler = Arc::new(Mutex::new(handler));
+
```
shannynalayna, line 71: Unnecessary change here

sdk/src/batches/store/diesel/mod.rs
```diff
@@ -192,3 +193,109 @@ impl BatchStore for DieselBatchStore<diesel::sqlite::SqliteConnection> {
         .relinquish_claim(id)
     }
 }
+
+pub struct DieselConnectionBatchStore<'a, C>
+where
+    C: diesel::Connection<TransactionManager = AnsiTransactionManager> + 'static,
```
peterschwarz, line 199: Do the constraints need to be this explicit in these structs?

daemon/src/database/mod.rs
dnewh: updated

daemon/src/splinter/app_auth_handler/mod.rs
dnewh: updated

daemon/src/database/mod.rs
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Cargill Incorporated
+ * Copyright 2019-2021 Cargill Incorporated
```
peterschwarz, line 2: Typo in commit message: InContextStoreFactorys -> InContextStoreFactory

daemon/src/event/db_handler.rs
```diff
@@ -108,388 +103,454 @@ use super::{CommitEvent, EventError, EventHandler, StateChange, IGNORED_NAMESPAC
...
-            tnt_store,
-        }
+        Self { store_factory }
     }
 }
 
-#[cfg(feature = "database-postgres")]
-impl EventHandler for DatabaseEventHandler<diesel::pg::PgConnection> {
+impl DatabaseEventHandler {
+    pub fn from_store_factory(store_factory: Box<dyn TransactionalStoreFactory>) -> Self {
```
peterschwarz, line 119: This could probably just be called `new`

daemon/src/event/db_handler.rs
```diff
@@ -108,388 +103,454 @@ use super::{CommitEvent, EventError, EventHandler, StateChange, IGNORED_NAMESPAC
...
+
+        trace!("The following operations will be performed: {:#?}", db_ops);
+
+        match txn
+            .get_grid_commit_store()
+            .get_commit_by_commit_num(commit.commit_num)
+        {
+            Ok(Some(ref b)) if b.commit_id != commit.commit_id => {
+                txn.get_grid_commit_store()
+                    .resolve_fork(commit.commit_num)?;
```
peterschwarz, line 159: These errors are missed, and would cause the state of the transaction to be neither rolled back nor committed.  The whole block of code that makes use of the transaction should be surrounded in a closure that returns `Result<(), EventError>` and then that result can be evaluated and the transaction committed or rolled back.

peterschwarz (changes requested):

sdk/src/batches/store/diesel/mod.rs
dnewh: After discussing, leaving this due to nested transactions.

daemon/src/database/mod.rs
dnewh: updated

daemon/src/event/db_handler.rs
dnewh: updated

daemon/src/event/db_handler.rs
dnewh: updated

peterschwarz (approved):

peterschwarz (approved):

PR #901, opened 2021-08-17
Add initial implementation of the "create purchase order" action

shannynalayna:
This PR makes a few updates to the purchase order protos and corresponding protocol code. This also adds an initial implementation of the "create purchase order" action for the Purchase Order smart contract's transaction handler.

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -27,16 +27,21 @@ cfg_if! {
     }
 }
 
-use crate::state::PurchaseOrderState;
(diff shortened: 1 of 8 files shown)


contracts/purchase_order/src/handler.rs
```diff
@@ -155,3 +252,22 @@ fn update_version(
 ) -> Result<(), ApplyError> {
     unimplemented!();
 }
+
```
dnewh, line 255: This will be removed in favor of the workflow-aware version. LGTM

dnewh (approved):

PR #902, opened 2021-08-19
Add workflow-aware has_permission check

dnewh:
This adds a new version of the has_permission check that is used in
other smart contracts. This check is workflow-aware by checking against
workflow permissions and transition permissions. This is used in the
same way as the original has_permission check. Agents will need to have
Pike roles with permissions that map to the names of Workflow permission
aliases for the given workflow.


contracts/purchase_order/src/handler.rs
```diff
@@ -155,3 +156,487 @@ fn update_version(
...
+#[allow(dead_code)]
+fn check_permission_with_workflow(
+    context: &mut dyn TransactionContext,
+    permission: &str,
+    signer: &str,
+    record_owner: &str,
+    workflow_state: WorkflowState,
+    desired_state: &str,
+) -> Result<bool, ApplyError> {
+    let perm_checker = PermissionChecker::new(context);
```
shannynalayna, line 169: Can the `PermissionChecker` just be passed to this function? This would remove the need to pass along the context since this struct is already being created before the payloads are passed onto the actions

contracts/purchase_order/src/handler.rs
```diff
@@ -155,3 +156,487 @@ fn update_version(
 ) -> Result<(), ApplyError> {
     unimplemented!();
 }
+
+#[allow(dead_code)]
+fn check_permission_with_workflow(
+    context: &mut dyn TransactionContext,
+    permission: &str,
```
shannynalayna, line 163: Would it make sense to have a similar function to `permission_to_perm_string` defined for the actions so this can use some sort of enum rather than just a string defined in each action? Doesn't change the impl here, more general question based on this. I'm not sure all of the permissions that would need to be included, tho (just `can-create/update-po` and `can-create/update-version`?)

contracts/purchase_order/src/handler.rs
dnewh: sure, updated

contracts/purchase_order/src/handler.rs
dnewh: It definitely could, it will definitely be a longer list of perms than other modules but that's probably even more reason to have it enumerated

contracts/purchase_order/src/handler.rs
```diff
@@ -155,3 +156,490 @@ fn update_version(
 ) -> Result<(), ApplyError> {
     unimplemented!();
 }
+
+#[allow(dead_code)]
+fn check_permission_with_workflow(
+    perm_checker: PermissionChecker,
```
shannynalayna, line 162: You should make this a reference as it doesn't need ownership and would be cloned from a ref anyway

dnewh: :+1:

contracts/purchase_order/src/handler.rs
```diff
@@ -155,3 +156,490 @@ fn update_version(
...
+
+#[allow(dead_code)]
+fn check_permission_with_workflow(
+    perm_checker: &PermissionChecker,
+    permission: &str,
+    signer: &str,
+    record_owner: &str,
+    workflow_state: WorkflowState,
+    desired_state: &str,
+) -> Result<bool, ApplyError> {
```
shannynalayna, line 168: Seems simpler to return just `()` here rather than a bool that would need to be parsed. The case that the bool is still false after running through the aliases, this method should probably return an invalid transaction err that the signer doesn't have the correct permission.

shannynalayna (approved):

contracts/purchase_order/src/handler.rs
dnewh: updated

PR #903, opened 2021-08-19
Implement some methods for Purchase Order state

shannynalayna:
This PR adds implementations for a few methods on the Purchase Order state that are used by the smart contract actions. This pr adds implementations for the methods `get_purchase_order`, `set_purchase_order`, `get_agent`, and `get_organization`.

### DIFF
--- a/contracts/purchase_order/src/state.rs
+++ b/contracts/purchase_order/src/state.rs
@@ -25,9 +25,9 @@ cfg_if! {
 use grid_sdk::{
     protocol::{
         pike::state::{Agent, Organization},
-        purchase_order::state::{PurchaseOrder, PurchaseOrderList},
+        purchase_order::state::{PurchaseOrder, PurchaseOrderList, PurchaseOrderListBuilder},
     },
-    protos::FromBytes,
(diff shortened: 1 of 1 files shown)


dnewh (approved):

contracts/purchase_order/src/state.rs
```diff
@@ -130,7 +130,22 @@ impl<'a> PurchaseOrderState<'a> {
         }
     }
 
-    pub fn _get_organization(&self, _id: &str) -> Result<Option<Organization>, ApplyError> {
-        unimplemented!();
+    pub fn _get_organization(&self, id: &str) -> Result<Option<Organization>, ApplyError> {
```
dplumb94, line 133: I still feel like we should but these in the SDK at some point so we dont have to rewrite them in every contract.

dplumb94 (approved):
LGTM

contracts/purchase_order/src/state.rs
shannynalayna: I made an issue in the backlog for this: https://github.com/hyperledger/grid/issues/905

PR #912, opened 2021-08-26
Add workflow-aware "Create PO" action

shannynalayna:
This change adds an implementation to the purchase order smart contract's "create po" action. This action is implemented using the workflow-aware permission check.

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -135,19 +135,19 @@ fn create_purchase_order(
     let buyer_org_id = payload.buyer_org_id().to_string();
     let seller_org_id = payload.seller_org_id().to_string();
     // Check that the organizations owning the purchase order exist
-    state._get_organization(&buyer_org_id)?.ok_or_else(|| {
+    state.get_organization(&buyer_org_id)?.ok_or_else(|| {
         ApplyError::InvalidTransaction(format!("Organization {} does not exist", &buyer_org_id))
     })?;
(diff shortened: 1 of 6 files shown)


dnewh (approved):

dnewh (dismissed):

contracts/purchase_order/src/handler.rs
```diff
@@ -121,12 +127,188 @@ impl TransactionHandler for PurchaseOrderTransactionHandler {
...
+                    beginning_workflow.subworkflow("version").ok_or_else(|| {
+                        ApplyError::InternalError("Unable to get `version` subworkflow".to_string())
+                    })?;
+                let start_state = version_subworkflow.state("editable").ok_or_else(|| {
+                    ApplyError::InternalError(
+                        "Unable to get create state from subworkflow".to_string(),
+                    )
+                })?;
+                let perm_result = perm_checker
+                    ._check_permission_with_workflow(
```
dnewh, line 193: This should probably also get rid of the "_" on the `check_permission_with_workflow` fn now that it's being used

dnewh (changes requested):

contracts/purchase_order/src/handler.rs
shannynalayna: Updated!

dnewh (approved):

dnewh (approved):

leebradley (approved):

KevinAnsleyJohnson (approved):

agunde406 (approved):

PR #915, opened 2021-08-31
Add purchase order list and show functionality

KevinAnsleyJohnson:
Added purchase order list and show functionality in the CLI. This 
included updating the SDK client's fetch_entities_list to support 
multiple query string parameters via a hashmap, implementing list and 
show for purchase orders within ReqwestPurchaseOrderClient, and 
implementing the list and show functionality for the CLI.

### DIFF
--- a/cli/src/actions/purchase_orders.rs
+++ b/cli/src/actions/purchase_orders.rs
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::cmp;
 use std::convert::TryInto;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use grid_sdk::{
-    client::purchase_order::{PurchaseOrderClient, PurchaseOrderRevision},
+    client::purchase_order::{
+        PurchaseOrder, PurchaseOrderClient, PurchaseOrderFilter, PurchaseOrderRevision,
+        PurchaseOrderVersion,
+    },
     protocol::purchase_order::payload::{
         Action, CreatePurchaseOrderPayload, CreateVersionPayload, PurchaseOrderPayloadBuilder,
     },
@@ -24,8 +28,10 @@ use grid_sdk::{
     purchase_order::addressing::GRID_PURCHASE_ORDER_NAMESPACE,
 };
 
+use chrono::{DateTime, Utc};
 use cylinder::Signer;
 use rand::{distributions::Alphanumeric, Rng};
+use serde::Serialize;
 
 use crate::error::CliError;
 use crate::transaction::purchase_order_batch_builder;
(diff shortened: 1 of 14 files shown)


cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        Ok(())
+    }
+}
+
+fn print_formattable<T: std::fmt::Display + Serialize>(object: T, format: Option<&str>) {
+    match format {
+        Some("json") => {
+            let formatted = serde_json::to_string(&object)
+                .unwrap_or_else(|_| "Cannot Format Purchase Order".to_string());
```
KevinAnsleyJohnson, line 161: This maybe should match and choose to print an err!() vs a println!(). Thoughts?

cli/src/main.rs
```diff
@@ -2494,22 +2498,59 @@ fn run() -> Result<(), CliError> {
...
+            let purchase_order_client = client_factory.get_purchase_order_client(url);
+
             match m.subcommand() {
                 ("create", Some(_)) => unimplemented!(),
                 ("update", Some(_)) => unimplemented!(),
-                ("list", Some(_)) => unimplemented!(),
-                ("show", Some(_)) => unimplemented!(),
+                ("list", Some(_)) => {
+                    let filter = PurchaseOrderFilter {
+                        is_closed: if m.is_present("closed") {
```
KevinAnsleyJohnson, line 2519: I chose against a builder here because every field is optional. Not sure if there's a cleaner way to do it.

sdk/src/client/reqwest.rs
```diff
@@ -130,7 +130,6 @@ pub fn fetch_entities_list<T: DeserializeOwned>(
 ) -> Result<Vec<T>, ClientError> {
     let client = BlockingClient::new();
     let mut final_url = format!("{}/{}", url, route);
-
```
shannynalayna: This change looks unrelated to this commit

sdk/src/client/purchase_order/reqwest.rs
```diff
@@ -142,23 +165,48 @@ impl PurchaseOrderClient for ReqwestPurchaseOrderClient {
...
+        if let Some(filter) = filter {
+            if let Some(org_id) = filter.org_id {
+                org_id_placeholder = org_id;
+                filter_map.insert("org_id", org_id_placeholder.as_str());
+            }
+            // Note: This filter is not currently implemented in rest api
+            if let Some(is_closed) = filter.is_closed {
+                filter_map.insert("is_closed", boolean_to_str(&is_closed));
+            }
+            // Note: This filter is not currently implemented in rest api
```
shannynalayna, line 190: I think for now we should remove these notes and you should add a story to implement the filters in the po rest api - as it will be difficult to keep track of these comments to be removed when it is implemented

sdk/src/client/purchase_order/reqwest.rs
```diff
@@ -101,6 +107,13 @@ impl ReqwestPurchaseOrderClient {
     }
 }
 
+fn boolean_to_str(value: &bool) -> &'static str {
```
shannynalayna, line 110: I think you are able to use `to_string()` on a bool value in rust, so you could just call `&is_closed.to_string()` rather than implementing this separate function

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
+use std::time::SystemTime;
```
shannynalayna, line 1: Missing a copyright on this file

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+                    )?;
+                    write!(f, "\n\n")?;
+
+                    if let Some(current_revision) = current_version
+                        .revisions
+                        .iter()
+                        .find(|r| r.revision_id == current_version.current_revision_id)
+                    {
+                        write!(f, "\nRevision {}", current_revision.revision_id)?;
+                        // Note: Summary fields for order_xml_3_4 have not been decided yet.
```
shannynalayna, line 141: Again, I think this should become a story and removed as a note from the code

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+                    } else {
+                        write!(f, "\nRevision N/A")?;
+                    }
+                }
+            }
+            None => {
```
shannynalayna, line 147: If there is no accepted version, we should just not write out anything, I think

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+
+                    if let Some(current_revision) = current_version
+                        .revisions
+                        .iter()
+                        .find(|r| r.revision_id == current_version.current_revision_id)
+                    {
+                        write!(f, "\nRevision {}", current_revision.revision_id)?;
+                        // Note: Summary fields for order_xml_3_4 have not been decided yet.
+                    } else {
+                        write!(f, "\nRevision N/A")?;
```
shannynalayna, line 143: Same here, if the revision doesn't exist, I don't think we should print anything out

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+            write!(f, "\n\n{}", purchase_order)?;
+        }
+        Ok(())
+    }
+}
+
+impl std::fmt::Display for PurchaseOrderCli {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Purchase Order:")?;
+        write!(f, "\n\t{:18}{}", "organization", &self.org_id)?;
```
shannynalayna, line 101: Hm, so after looking at how this writes out, I think it would make more sense if it were formatted like this: 
```
Purchase Order <UUID>:
        Organization: <Org ID>
        Workflow Status: <workflow status>
        Created At: <created at timestamp>
        Closed: <is_closed bool>
```
And then we can check if there is the accepted version and print it in a similar format

cli/src/actions/purchase_orders.rs
```diff
@@ -0,0 +1,205 @@
...
+        write!(f, "\n\t{:18}{}", "uuid", &self.uuid)?;
+        // Note: purchase order number is not in payload today.
```
shannynalayna, line 103: We should also change (across this entire PR, really) `uuid` to `uid`

shannynalayna (changes requested):

cli/src/main.rs
shannynalayna: I think based on how the flags are being used, I am not sure I can think of a cleaner way to this either really

cli/src/actions/purchase_orders.rs
shannynalayna: I think this function should return a `CliError`, so it is logged and output to the user rather than just output to the user

sdk/src/client/purchase_order/reqwest.rs
KevinAnsleyJohnson: to_string() works for a String, but I can't to .to_string().as_str() because the temporary is freed while still in use... that's why I introduced the static str method. Is there a different way to do this?

cli/src/actions/purchase_orders.rs
KevinAnsleyJohnson: fixed in latest version!

sdk/src/client/purchase_order/reqwest.rs
shannynalayna: I think in this case, we can put a `&` in front of `is_closed.to_string()` and that will be the right type to add it to the hashmap. So it would look like this when you add the bool value to a hashmap: 
`filter_map.insert("is_closed", &is_closed.to_string());`

KevinAnsleyJohnson: Unfortunately, it still causes a temporary is freed while still in use... this is due to the `If let Some` statement.

shannynalayna (approved):

dnewh (approved):

leebradley (approved):

shannynalayna:
This PR is being closed in favor of this PR: https://github.com/hyperledger/grid/pull/1008, which is more up-to-date with the current po structures

PR #939, opened 2021-09-17
Hookup PO SDE

dnewh:
This sets up the state delta export for purchase order events. This PR also adds the conversion methods to `EventError` for the purchase order store errors.

### DIFF
--- a/daemon/src/event/db_handler.rs
+++ b/daemon/src/event/db_handler.rs
@@ -64,6 +64,21 @@ use grid_sdk::{
     },
     protocol::product::state::ProductList,
 };
+#[cfg(feature = "purchase-order")]
+use grid_sdk::{
+    protocol::purchase_order::state::{
+        PurchaseOrderList, PurchaseOrderRevision as ProtocolPORevision,
+        PurchaseOrderVersion as ProtocolPOVersion,
+    },
+    purchase_order::{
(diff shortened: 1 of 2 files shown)


daemon/src/event/db_handler.rs
```diff
@@ -942,3 +1007,64 @@ fn make_location_attributes(
...
+
+#[cfg(feature = "purchase-order")]
+fn make_po_revisions(
+    revisions: Vec<ProtocolPORevision>,
+    start_commit_num: i64,
+    service_id: Option<&String>,
+) -> Result<Vec<PurchaseOrderVersionRevision>, EventError> {
+    let mut revs = Vec::new();
+
+    for revision in revisions {
```
shannynalayna, line 1049: You could probably simplify these functions to iterating over the `revisions` list with `.for_each` and converting the objects to the `PurchaseOrderVersionRevision` struct, rather than creating a new list.

shannynalayna: Same thing here goes for the `make_po_versions` func ^

shannynalayna: Realize now there's a fn being used within this builder that will make it not easy to throw this in a for_each loop

shannynalayna (dismissed):

leebradley (approved):

shannynalayna (approved):

KevinAnsleyJohnson (approved):

vaporos (approved):

PR #954, opened 2021-09-24
Add list/get revision operations

dnewh:
This adds database operations to the `PurchaseOrderStore` to get and list `PurchaseOrderVersionRevisions`. The list operation lists revisions filtered by `po_uid` and `version_id` and the get operation fetches a revision filtered by `po_uid`, `version_id`, and `revision_id`. This also includes migrations and updates to the revisions table to add a `purchase_order_uid` column. This is necessary to be able to properly get revisions in case of overlap in version IDs between different POs.

### DIFF
--- a/sdk/src/purchase_order/store/mod.rs
+++ b/sdk/src/purchase_order/store/mod.rs
@@ -838,6 +838,22 @@ pub trait PurchaseOrderStore {
         service_id: Option<&str>,
     ) -> Result<Option<PurchaseOrderVersion>, PurchaseOrderStoreError>;
 
+    /// Fetches a purchase order revision from the underlying storage
+    ///
+    /// # Arguments
+    ///
+    ///  * `po_uid`    - The uid of the purchase order the revision belongs to
+    ///  * `version_id` - The ID of the version the revision is for
+    ///  * `revision_id` - The ID of the revision to fetch
+    ///  * `service_id` - The service ID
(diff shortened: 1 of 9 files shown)


sdk/src/migrations/diesel/postgres/migrations/2021-09-24-150300_update_po_revisions_table/down.sql
```diff
@@ -0,0 +1,16 @@
+-- Copyright 2021 Cargill Incorporated
```
shannynalayna, line 1: I think we can remove the additional migration data for this update, so update the existing migration data (as this is still unreleased).

dnewh: updated

sdk/src/purchase_order/store/mod.rs
```diff
@@ -447,6 +447,19 @@ impl PurchaseOrderVersionBuilder {
     }
 }
 
+/// Represents a list of Grid Purchase Orders
```
shannynalayna, line 450: Can we update this comment to specify this is a list of revisions?

shannynalayna (approved):

leebradley (approved):

agunde406 (approved):

PR #955, opened 2021-09-27
Make `accepted_version_id` optional throughout SDK

shannynalayna:
This updates the `accepted_version_id` field in the purchase order object to be optional throughout the SDK. This previously was defined as a required string, however this value may not always be defined. A purchase order version must first be validated that it is able to be "accepted". Therefore, this value should be optional to allow for the cases where a purchase order does not have an accepted version (yet).


contracts/purchase_order/src/workflow.rs
```diff
@@ -398,3 +400,12 @@ fn collaborative_sub_workflow() -> SubWorkflow {
         .add_starting_state("create")
         .build()
 }
+
+impl fmt::Display for POWorkflow {
```
peterschwarz, line 404: This seems like an unrelated change

shannynalayna: Removed

peterschwarz (approved):

dnewh (approved):

PR #957, opened 2021-09-28
Add purchase order version list endpoint

leebradley:
Add an endpoint to obtain purchase order versions at
/purchase-order/{UUID}/versions.

Related: #866

### DIFF
--- a/sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
+++ b/sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
@@ -92,14 +92,32 @@ pub async fn get_purchase_order(
 
 #[get("/purchase-order/{uuid}/versions")]
 pub async fn list_purchase_order_versions(
-    _store_state: web::Data<StoreState>,
-    _uuid: web::Path<String>,
-    _query: web::Query<QueryServiceId>,
+    store_state: web::Data<StoreState>,
+    uuid: web::Path<String>,
+    query_service_id: web::Query<QueryServiceId>,
+    query_paging: web::Query<QueryPaging>,
     version: ProtocolVersion,
     _: AcceptServiceIdParam,
 ) -> HttpResponse {
+    let store = store_state.store_factory.get_grid_purchase_order_store();
     match version {
-        ProtocolVersion::V1 => unimplemented!(),
+        ProtocolVersion::V1 => {
+            let paging = query_paging.into_inner();
+            match v1::list_purchase_order_versions(
+                store,
+                uuid.into_inner(),
+                query_service_id.into_inner().service_id.as_deref(),
+                paging.offset(),
+                paging.limit(),
+            ) {
+                Ok(res) => HttpResponse::Ok().json(res),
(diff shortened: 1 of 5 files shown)


sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
```diff
@@ -92,14 +92,28 @@ pub async fn get_purchase_order(
 
 #[get("/purchase-order/{uuid}/versions")]
 pub async fn list_purchase_order_versions(
-    _store_state: web::Data<StoreState>,
-    _uuid: web::Path<String>,
-    _query: web::Query<QueryServiceId>,
+    store_state: web::Data<StoreState>,
+    uuid: web::Path<String>,
```
shannynalayna, line 96: Can you update this argument to be `uid` as the purchase order id was previously updated from `uuid` -> `uid` (throughout other components in Grid)

shannynalayna: Including the openapi spec, we can make these updates as you are implementing each endpoint. But it might be worthwhile as apart of these tasks to verify the openapi spec is up to date

sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
```diff
@@ -92,14 +92,28 @@ pub async fn get_purchase_order(
...
+    let store = store_state.store_factory.get_grid_purchase_order_store();
     match version {
-        ProtocolVersion::V1 => unimplemented!(),
+        ProtocolVersion::V1 => {
+            match v1::get_purchase_order(
+                store,
+                uuid.into_inner(),
+                query_service_id.into_inner().service_id.as_deref(),
+            ) {
+                Ok(res) => HttpResponse::Ok().json(res.version_numbers),
```
shannynalayna, line 109: It looks like this will return just the version ids for the specified PO, but we want to be returning a list of the `PurchaseOrderVersion` objects. Could we use the store's `list_purchase_order_versions` here? You can specify the purchase order's ID & service id for that op

shannynalayna (dismissed):

sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
```diff
@@ -90,16 +90,34 @@ pub async fn get_purchase_order(
     }
 }
 
-#[get("/purchase-order/{uuid}/versions")]
+#[get("/purchase-order/{uid}/versions")]
```
dnewh, line 93: Should this be `.../version` to match the other routes?

dnewh (changes requested):

sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
leebradley: I agree. Updated the endpoints and spec so that all these version endpoints are singular.

dnewh (approved):

shannynalayna (approved):

PR #959, opened 2021-09-28
Add OpenAPI validation

leebradley:
Add a GitHub Action to validate the swagger file


leebradley:
This PR updates CI to fail when there are validation problems with `daemon/openapi.yaml`.

There are currently validation problems, so this test will (correctly) fail.

vaporos:
This is cool, but any linting, we need to be able to replicate w/just as well; maybe "just lint-openapi" if it has to use docker.

leebradley:
Once https://github.com/hyperledger/grid/pull/965 is merged the OpenAPI validation step in this PR should pass

leebradley:
> This is cool, but any linting, we need to be able to replicate w/just as well; maybe "just lint-openapi" if it has to use docker.

Updated to add a command `just lint-openapi` that is now run by the GitHub actions

PR #960, opened 2021-09-29
Add CLI filters to DB ops

dnewh:
This adds filters to some of the PO DB operations that correspond to the filters available in the list/show PO CLI commands.


cli/man/grid-po-version-list.1.md
```diff
@@ -13,13 +13,19 @@ NAME
...
 
-**grid po version list** \[**FLAGS**\] \[**OPTIONS**\]
+**grid po version list** \[**FLAGS**\] \[**OPTIONS**\] <PURCHASE_ORDER_ID>
 
 DESCRIPTION
 ===========
 
 List purchase order versions in grid.
 
+ARGS
```
shannynalayna, line 23: Can we update the man page in a separate commit? I think it would be worthwhile to separate out that change so it's easily viewable

cli/man/grid-po-list.1.md
```diff
@@ -57,10 +57,14 @@ OPTIONS
...
-  
-`--org`
-: Optionally, filter the purchase orders for the organization specified by
-  `ORG_ID`. 
+
+`--buyer-org`
+: Optionally, filter the purchase orders for the buyer organization specified
+  by `ORG_ID`.
+
+`--seller-org`
```
shannynalayna, line 65: Hm, I wonder if it would make sense to keep the `org` argument or something like `org_id` and make this value match against both the buyer and seller org value in the db op? Would it be worthwhile to return all pos an org is involved in (regardless of buyer or seller) ?

dnewh: I think splitting it up is fine. If we have both the buyer and seller, I think it makes the most sense to have filters for both too.

cli/man/grid-po-version-list.1.md
dnewh: updated

leebradley (approved):

shannynalayna (approved):

sdk/src/rest_api/actix_web_3/routes/purchase_orders.rs
```diff
@@ -37,11 +39,20 @@ pub struct QueryRevisionNumber {
...
+pub struct QueryIsOpen {
+    pub is_open: Option<bool>,
+}
+
 #[get("/purchase-order")]
 pub async fn list_purchase_orders(
     store_state: web::Data<StoreState>,
-    query_buyer_org_id: web::Query<QueryOrgId>,
-    query_seller_org_id: web::Query<QueryOrgId>,
+    query_filters: web::Query<ListPOFilters>,
```
shannynalayna, line 65: Are the `QueryIsOpen` and `QueryHasAcceptedVersion` still needed here since it looks like we're just using the `ListPOFilters` struct?

sdk/src/purchase_order/store/mod.rs
```diff
@@ -821,12 +827,16 @@ pub trait PurchaseOrderStore {
     /// # Arguments
     ///
     ///  * `po_uid`   - The uid of the purchase order to get versions for
+    ///  * `is_accepted - Optional filter to get just the accepted or not
+    ///    accepted versions for a PO
+    ///  * `is_draft` - Optional filter to get versions marked as draft or not
```
shannynalayna, line 832: Since we're consolidating the filters in the `list_purchase_orders` method, perhaps we should keep that consistent so making these one bullet point

dnewh: updated

PR #967, opened 2021-10-01
Add implementation for "create version" action in po smart contract

shannynalayna:
This PR mainly adds an implementation and unit tests for the "create_version" function within the purchase order smart contract's handler.


contracts/purchase_order/src/handler.rs
```diff
@@ -315,12 +315,120 @@ fn update_purchase_order(
...
+            ))
+        })?
+        .ok_or_else(|| {
+            ApplyError::InvalidTransaction(format!("Purchase order {} does not exist", po_uid))
+        })?
+        .workflow_type()
+        .to_string();
+    // Validate this version does not already exist
+    if state
+        ._get_purchase_order_version(po_uid, version_id)?
```
dnewh, line 348: This fn should be updated to remove the "_" as it is being used now

dnewh (changes requested):

contracts/purchase_order/src/handler.rs
shannynalayna: Added a commit to remove the underscore for the get/set po version state methods

dnewh (approved):

leebradley (approved):

PR #968, opened 2021-10-04
Create PO in cli

KevinAnsleyJohnson:
Adding functionality to create purchase order. Based on the updated RFC,
we need buyer-org and seller-org instead of owning org. Changes are
reflected in the man pages as well as code.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -47,6 +47,7 @@ serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.8"
 diesel_migrations = "1.4"
 whoami = "1"
+rand = { version = "0.7", optional = true }
 
 [dev-dependencies]
 tempdir = "0.3"
@@ -88,7 +89,7 @@ sqlite = [
 location = ["pike", "schema"]
 pike = []
 product = ["pike", "schema"]
-purchase-order = ["grid-sdk/purchase-order"]
+purchase-order = ["grid-sdk/purchase-order", "rand"]
 sawtooth = []
 schema = ["pike"]
 splinter = []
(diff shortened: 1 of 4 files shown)


cli/Cargo.toml
```diff
@@ -47,6 +47,7 @@ serde = { version = "1", features = ["derive"] }
 serde_yaml = "0.8"
 diesel_migrations = "1.4"
 whoami = "1"
+rand = "0.7"
```
dnewh, line 50: This should be pulled in by the purchase-order feature

dnewh (changes requested):

cli/src/main.rs
```diff
@@ -1431,9 +1433,19 @@ fn run() -> Result<(), CliError> {
...
-                            Arg::with_name("org")
-                                .value_name("org_id")
-                                .long("org")
+                            Arg::with_name("buyer_org_id")
+                                .value_name("buyer_org_id")
+                                .long("buyer-org")
+                                .short("buyer")
+                                .takes_value(true)
+                                .required(true)
+                                .help("ID of the organization which owns the Purchase Order"),
```
shannynalayna, line 1442: Perhaps we could change this "help" text to be more specific for the buyer and seller org options

leebradley (approved):

cli/Cargo.toml
KevinAnsleyJohnson: fixed!

cli/src/main.rs
KevinAnsleyJohnson: fixed!

shannynalayna (approved):

dnewh (approved):

PR #978, opened 2021-10-11
Update "create po" smart contract action

shannynalayna:
This change updates the "create po" action implementation to handle the `workflow_status` now present in the `CreatePurchaseOrderPayload` and the `CreateVersionPayload`. 

Previously, this action assumed the intended `workflow` status based on the state of the payload. Now, the purchase order is created with the intended `workflow_status` in mind and this transition is validated rather than what the action assumes the desired workflow state is.

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -724,17 +724,31 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


shannynalayna:
Based on our updated understanding of how workflow constraints will interact with the smart contract, this PR will be closed in favor of another approach that utilizes the compartmentalized workflow constraints.

PR #1006, opened 2021-10-29
Update schema loading for Order.xsd

shannynalayna:
This change updates the loading of the schema to be done in the
directory where schemas are specified. This directory is able to be set
with the `GRID_ORDER_SCHEMA_DIR` environment variable, or it is set by
default.

### DIFF
--- a/sdk/src/data_validation/xml/xml_ffi.rs
+++ b/sdk/src/data_validation/xml/xml_ffi.rs
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 use core::marker::{PhantomData, PhantomPinned};
+use std::env;
 use std::ffi::CString;
 
+use super::Schema;
 use crate::error::InternalError;
 
 use libc::{c_char, c_int, c_uint};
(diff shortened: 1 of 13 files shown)


leebradley (dismissed):
Tested locally. Correctly validates po XML

dnewh (dismissed):

sdk/src/data_validation/xml/xml_ffi.rs
```diff
@@ -33,8 +38,33 @@ struct XmlSchemaPtr {
...
+        schema_type: &Schema,
+    ) -> Result<Self, InternalError> {
+        let schema_ptr = match schema_type {
+            Schema::OrderXmlV3_4 => {
+                // Check if the `GRID_SCHEMA_DIR` environment variable is set, which tells Grid where
+                // to locate any dependent schema files
+                let schema_dir = match env::var(GRID_ORDER_SCHEMA_DIR).ok() {
+                    Some(schema_path) => PathBuf::from(schema_path),
+                    None => {
+                        let mut base_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
peterschwarz, line 52: This only works when running from the source directory. It doesn't seem like a good default value.

shannynalayna: I updated these methods to simply take in a directory argument. This allows for the unit tests to set the directory relative to the `CARGO_MANIFEST_DIR` env var (which originally was a case this match arm covered), and it also allows an end user to set these values using an environment variable (with a default, which directs to the sdk's `data_validation` module). I'm not sure if there's a better default in this case. And, this will also allow us to point to a new directory, meaning we could copy over the schema files in a docker env in the near future and point these variables to that directory.

dnewh (dismissed):

leebradley (dismissed):

dnewh (approved):

peterschwarz (approved):

dnewh (approved):

Issue #985, opened 2021-10-18
Update "Create PO" action with updated workflows and payload


shannynalayna:
Currently, the Create po payloads have the additional `workflow_status` and `alternate-ids` fields. The built-in workflows have also been updated to add constraints. 

The create po action implementation does not consider the workflow status, constraints, or alternate IDs available in the payload. This action must be implemented to consider these additional fields.

PR #1009, opened 2021-11-01
Update po create action

leebradley:
Addresses https://github.com/hyperledger/grid/issues/985

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -271,6 +271,7 @@ fn create_purchase_order(
         .with_uid(po_uid.to_string())
         .with_versions(versions)
         .with_workflow_status(payload.workflow_status().to_string())
+        .with_alternate_ids(payload.alternate_ids().to_vec())
         .with_created_at(payload.created_at())
         .with_is_closed(false)
         .with_buyer_org_id(buyer_org_id)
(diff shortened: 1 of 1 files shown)


contracts/purchase_order/src/handler.rs
```diff
@@ -153,11 +153,49 @@ fn create_purchase_order(
...
+                        agent.org_id(),
+                        payload_version.workflow_status()
+                    )));
+            }
+
+            let payload_revision = payload_version.revision();
+
             let revision = PurchaseOrderRevisionBuilder::new()
                 .with_revision_id(payload_revision.revision_id())
                 .with_submitter(signer.to_string())
```
shannynalayna, line 207: I've been thinking about this field, and I think it should be `payload_revision.submitter()` rather than taking the signer directly.

contracts/purchase_order/src/handler.rs
```diff
@@ -27,6 +27,7 @@ cfg_if! {
     }
 }
 
+use super::payload::validate_create_po_payload;
```
shannynalayna, line 30: Once #1017 is merged, this commit will be unnecessary, though I think this is a non-blocking comment if this one gets approved first, then 1017 should update this func to remove the redundant validation

leebradley: @shannynalayna Cool. Updated to remove the commits that added validation

shannynalayna (approved):

dnewh (approved):

PR #1016, opened 2021-11-03
Update `add` method in state to consider updates

shannynalayna:
This change updates the `set_purchase_order_version` method in the
purchase order state module to consider if a purchase order version is
being updated. This means, when the po is fetched from state, the method
will now check to see if a version with the same ID exists in the list
and then it is removed and re-inserted with the updated version.
Otherwise, a new version is simply added to the end of the purchase
order's `versions` list.

### DIFF
--- a/contracts/purchase_order/src/state.rs
+++ b/contracts/purchase_order/src/state.rs
@@ -183,9 +183,26 @@ impl<'a> PurchaseOrderState<'a> {
                 po_uid
             ))),
         }?;
+
         // Add the `PurchaseOrderVersion` to the purchase order
         let mut versions = orig_po.versions().to_vec();
-        versions.push(purchase_order_version);
+        let vers_index = versions
+            .iter()
+            .position(|vers| vers.version_id() == purchase_order_version.version_id());
+        match vers_index {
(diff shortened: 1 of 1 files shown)


contracts/purchase_order/src/state.rs
```diff
@@ -185,7 +185,27 @@ impl<'a> PurchaseOrderState<'a> {
         }?;
         // Add the `PurchaseOrderVersion` to the purchase order
         let mut versions = orig_po.versions().to_vec();
-        versions.push(purchase_order_version);
+        let mut vers_index = None;
```
dnewh, line 188: Like @leebradley mentioned, this could use `.position` to avoid making it mutable

dnewh (changes requested):

contracts/purchase_order/src/state.rs
shannynalayna: Alright I updated that part to use position

leebradley (approved):

dnewh (approved):

agunde406 (approved):

PR #1030, opened 2021-11-08
Support `alternate_id` for PO creation and lookup

dnewh:
This updates the Grid PO store, SDE, and CLI to support PO creation and lookup using a properly formatted (`<id_type>:<id>`) alternate ID. This change includes an update to the `grid create po` CLI command to update the flag for specifying an alternate ID to `--alternate-id` from the less descriptive `--id`.

### DIFF
--- a/cli/man/grid-po-create.1.md
+++ b/cli/man/grid-po-create.1.md
@@ -42,7 +42,7 @@ output.
 OPTIONS
 =======
 
-`--id`
+`--alternate-id`
 : Optionally include an alternate ID. This may be specified multiple times.
 An ID is of the format `alternate_id_type:alternate_id`.
 Examples: `po_number:12348959` and/or `internal_po_id:a8f9fke`.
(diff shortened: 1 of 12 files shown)


cli/man/grid-po-create.1.md
```diff
@@ -42,7 +42,7 @@ output.
 OPTIONS
 =======
 
-`--id`
```
leebradley, line 45: We should update the example on line 86 to also use `--alternate-id`.

leebradley (changes requested):
Tested locally. One minor change, otherwise LGTM.

cli/man/grid-po-create.1.md
dnewh: updated

leebradley (approved):

shannynalayna (approved):
Tested

agunde406 (approved):

PR #1038, opened 2021-11-11
Fix po update cli command

leebradley:
This change fixes the po update cli command by changing it to use the alternate id from the po get endpoint (which is already called earlier and therefore its result is available) instead of the unimplemented `list_alternate_ids` REST client fn.


shannynalayna (approved):
Tested

shannynalayna (approved):

dnewh (approved):

PR #1062, opened 2021-11-15
Add concern for workflow constraints to "create vers" action

shannynalayna:
This change adds concern for workflow constraints to the "create
version" action of the purchase order smart contract. This also updates
the tests written for this action to re-align with the updated
implementation.

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -496,7 +496,7 @@ fn create_version(
     state: &mut PurchaseOrderState,
     perm_checker: &PermissionChecker,
 ) -> Result<(), ApplyError> {
-    // Validate the signer exists as an agent
+    // Validate the signer exists as an agent and retrieve the agent's organization ID
     let agent_org_id = state
         .get_agent(signer)?
         .ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


contracts/purchase_order/src/handler.rs
```diff
@@ -529,40 +523,63 @@ fn create_version(
...
+    // Check the desired workflow state for any constraints it may have related to the version
+    if desired_workflow_state.has_constraint(&WorkflowConstraint::Accepted.to_string())
+        && payload.is_draft()
     {
-        return Err(ApplyError::InvalidTransaction(format!(
-            "Version draft status {} does not match intended workflow state {}",
-            payload.is_draft(),
-            &desired_state,
-        )));
+        return Err(ApplyError::InvalidTransaction(
```
shannynalayna, line 547: This should maybe also check the "accepted version number" of the associated po to validate the version is able to be moved into the "accepted" state

shannynalayna: But, I'm not sure if a purchase order can "accept" a version before it has been created

dnewh: I think it's fine this way b/c you have to create a PO before the version anyway

dnewh (approved):

leebradley (approved):

vaporos (approved):

PR #1068, opened 2021-11-16
Enforce uniqueness for PO alternate IDs

dnewh:
This updates the purchase order code to enforce uniqueness for alternate IDs. This adds checks to both the client and PO store to make sure that alternate IDs are only used once per splinter circuit or sawtooth instance.

### DIFF
--- a/sdk/src/client/reqwest/purchase_order/mod.rs
+++ b/sdk/src/client/reqwest/purchase_order/mod.rs
@@ -133,6 +133,9 @@ impl PurchaseOrderClient for ReqwestPurchaseOrderClient {
             if let Some(seller_org_id) = filters.seller_org_id {
                 filter_map.insert("seller_org_id", seller_org_id);
             }
+            if let Some(alternate_ids) = filters.alternate_ids {
+                filter_map.insert("alternate_ids", alternate_ids);
+            }
         }
         let dto_vec = fetch_entities_list::<data::PurchaseOrder>(
             &self.url,
--- a/sdk/src/purchase_order/store/diesel/operations/list_purchase_orders.rs
+++ b/sdk/src/purchase_order/store/diesel/operations/list_purchase_orders.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::PurchaseOrderStoreOperations;
+use super::{get_uid_from_alternate_id, PurchaseOrderStoreOperations};
 use crate::commits::MAX_COMMIT_NUM;
 use crate::error::InternalError;
 use crate::paging::Paging;
@@ -30,6 +30,7 @@ use crate::purchase_order::store::diesel::{
 };
 
 use crate::purchase_order::store::PurchaseOrderStoreError;
+use diesel::dsl::*;
 use diesel::prelude::*;
 
 pub(in crate::purchase_order::store::diesel) trait PurchaseOrderStoreListPurchaseOrdersOperation {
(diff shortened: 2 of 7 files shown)


sdk/src/client/reqwest/purchase_order/mod.rs
```diff
@@ -188,4 +188,42 @@ impl PurchaseOrderClient for ReqwestPurchaseOrderClient {
...
+        &self,
+        ids: Vec<AlternateId>,
+        service_id: Option<&str>,
+    ) -> Result<bool, ClientError> {
+        let dto_vec = fetch_entities_list::<data::PurchaseOrder>(
+            &self.url,
+            PO_ROUTE.to_string(),
+            service_id,
+            None,
+        )?;
```
leebradley, line 203:
```rust
let dto_vec = fetch_entities_list::<data::PurchaseOrder>(
    &self.url,
    PO_ROUTE.to_string(),
    service_id,
    None,
)?;
```

We're indexing the server purchase orders as though we expect a large dataset. I don't think it makes sense to fetch that entire dataset in the client, and then iterate over items client-side multiple times to do unique id constraint checks.

There are a couple of alternatives to this, such as making a server endpoint that will return a list of purchase orders by a list of ids (this could possibly be implemented merely as a filter on the list po endpoint with a comma-delimited list of ids). We pass all the alternate ids to that endpoint. If the dataset is empty, the client can be reasonably sure the unique constraint is not violated.

Which kind of brings me to another thought, that `check_alternate_ids_are_unique` is maybe too specific a client function and could be replaced by something more generic like listing purchase orders by a list of ids.

leebradley (changes requested):

sdk/src/purchase_order/store/diesel/schema.rs
```diff
@@ -59,8 +59,7 @@ table! {
 }
 
 table! {
-    purchase_order_alternate_id (id) {
-        id -> Int8,
+    purchase_order_alternate_id (alternate_id_type, alternate_id, service_id) {
```
shannynalayna, line 62: Curious how this operates with a sawtooth backend, does it just not include that value in the composite key if service id is none ?

dnewh: yeah, that table still has the service_id column but the value is always `None` or however the db handles that value

sdk/src/client/reqwest/purchase_order/mod.rs
dnewh: updated to do the check in the cli/store

sdk/src/client/reqwest/purchase_order/mod.rs
```diff
@@ -188,4 +188,42 @@ impl PurchaseOrderClient for ReqwestPurchaseOrderClient {
 
         Ok(dto.iter().map(PurchaseOrderRevision::from).collect())
     }
+
+    /// Checks whether a list of alternate IDs contains any that already exist
+    fn check_alternate_ids_are_unique(
+        &self,
+        ids: Vec<AlternateId>,
+        service_id: Option<&str>,
+    ) -> Result<bool, ClientError> {
```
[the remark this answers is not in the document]
dnewh: updated

sdk/src/purchase_order/store/diesel/operations/list_purchase_orders.rs
```diff
@@ -59,15 +60,45 @@ impl<'a> PurchaseOrderStoreListPurchaseOrdersOperation
...
+
+            if let Some(ref alternate_ids) = alternate_ids {
+                let id_vec = alternate_ids
+                    .split(',')
+                    .map(String::from)
+                    .collect::<Vec<String>>();
+                uids = id_vec
+                    .iter()
+                    .filter_map(
+                        |id| match get_uid_from_alternate_id::pg::get_uid_from_alternate_id(
```
leebradley, line 76: We might want to change this in the future to avoid SQL queries within loops, but TBH this specific call is unlikely to become a performance issue unless we pass a very large number of alternate ids.

leebradley (approved):
Tested locally. LGTM!

shannynalayna (changes requested):
Looks like this happens when running on Sawtooth: 
```
gridd-alpha                        | ERROR [gridd::event] An error occurred while handling events: Event Error: null value in column "service_id" of relation "purchase_order_alternate_id" violates not-null constraint
```

So, in the case that service id field is null (sawtooth) this error occurs... wonder if we could do a conditional id for sawtooth v splinter

dnewh:
> Looks like this happens when running on Sawtooth:
> 
> ```
> gridd-alpha                        | ERROR [gridd::event] An error occurred while handling events: Event Error: null value in column "service_id" of relation "purchase_order_alternate_id" violates not-null constraint
> ```
> 
> So, in the case that service id field is null (sawtooth) this error occurs... wonder if we could do a conditional id for sawtooth v splinter

removed the db check for now. The check in the cli/store operation should suffice

shannynalayna (approved):
Tested locally and the sawtooth example worked 👍

leebradley (approved):

vaporos (approved):

PR #1159, opened 2021-12-02
Update po version create and update commands

dnewh:
This updates the `grid po version create` and `grid po version update`
cli commands to take the po number as the first positional argument
rather than a required flag. This seems more natural and is the pattern
used for the list and show version commands.

### DIFF
--- a/cli/man/grid-po-version-create.1.md
+++ b/cli/man/grid-po-version-create.1.md
@@ -13,7 +13,7 @@ NAME
 SYNOPSIS
 ========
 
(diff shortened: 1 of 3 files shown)


CLEckhardt:
In the docs, should we call it ORDER_ID instead of PO_NUMBER since it's not really a number? Or PO_ID?

shannynalayna (approved):
docs can be updated in a separate PR

leebradley (approved):
Tested locally. Verified that the command passes the po id to the batch command correctly.

agunde406 (approved):

PR #1162, opened 2021-12-02
fix internal errs

shannynalayna:


dnewh (approved):

dnewh (approved):

CLEckhardt (approved):
Tested and worked for me.

dnewh (approved):

contracts/purchase_order/src/handler.rs
```diff
@@ -169,19 +169,25 @@ fn create_purchase_order(
             // Check the version permissions
             let perm_string = Permission::CanCreatePoVersion.to_string();
             let beginning_workflow = get_workflow(&POWorkflow::SystemOfRecord.to_string())
-                .ok_or_else(|| ApplyError::InternalError("Cannot build PO Workflow".to_string()))?;
+                .ok_or_else(|| {
```
vaporos, line 172: These changes appear correct, but the reasoning given in the commit message is incorrect.

Returning an "InternalError" from a transaction means that the processing of the transaction was non-deterministic as some unexpected (system-level) error occurred. It is a signal to Splinter/Sawtooth to retry the transaction later, hoping the issue is resolved. Resolutions of this type of error would include system resource issues (tried to read state but the database was down) or deploying a new version of the smart contract (because it was buggy). Actual errors like this are expected to be extremely rare when processing smart contracts.

An "InvalidTransaction" means "the transaction that was submitted was not valid". So, for example, if an workflow state is given that doesn't exist, that is an "InvalidTransaction" because processing it is inconsistent with the list of built-in workflows available. Importantly, this is not actually an error as it is normal operation for a smart contract to determine if a transaction is valid or invalid.

vaporos (changes requested):

contracts/purchase_order/src/handler.rs
shannynalayna: Updated the first commit message to better explain what's going on there

leebradley (approved):

dnewh (approved):

CLEckhardt (approved):

vaporos (approved):

PR #1172, opened 2021-12-08
Stabilization updates for `workflow`

shannynalayna:
This adds a few more stabilization updates, including beefed up module lvl and rustdoc comments and an update to the `can_transition` method.

### DIFF
--- a/sdk/src/workflow/subworkflow.rs
+++ b/sdk/src/workflow/subworkflow.rs
@@ -12,15 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! An API for managing a subprocess of a workflow
+//! An API for managing a subprocess within a workflow, which contains the list of states involved
+//! in this subprocess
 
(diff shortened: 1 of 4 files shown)


leebradley (dismissed):

sdk/src/workflow/subworkflow.rs
```diff
@@ -31,6 +33,8 @@ impl SubWorkflow {
         &self.name
     }
 
+    /// Retreive a specific workflow state from this subworkflow. Returns `None` if the state
```
dnewh, line 36: s/retreive/retrieve

dnewh (changes requested):

sdk/src/workflow/subworkflow.rs
shannynalayna: Updated!

dnewh (approved):

leebradley (approved):

agunde406 (approved):

PR #1182, opened 2021-12-13
v0.2 Rust 1.57 clippy updates

shannynalayna:
Rust 1.57 clippy updates for the 0-2 branch

### DIFF
--- a/cli/src/actions/products.rs
+++ b/cli/src/actions/products.rs
@@ -68,7 +68,11 @@ pub struct GridPropertyValue {
     pub lat_long_value: Option<LatLong>,
 }
 
+// The `LatLong` value is used by the `GridPropertyValue` `lat_long_value` field and returned to
+// the user. Rust 1.57 does not recognize the fields of this struct as being read, therefore we
+// must use the `allow(dead_code)` decorator for this struct.
 #[derive(Debug, Deserialize, Clone)]
+#[allow(dead_code)]
 pub struct LatLong {
     latitude: i64,
     longitude: i64,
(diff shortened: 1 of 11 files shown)


dnewh (approved):

leebradley (approved):

cli/src/actions/products.rs
```diff
@@ -68,7 +68,11 @@ pub struct GridPropertyValue {
     pub lat_long_value: Option<LatLong>,
 }
 
+/// The `LatLong` value is used by the `GridPropertyValue` `lat_long_value` field and returned to
+/// the user. Rust 1.57 does not recognize the fields of this struct as being read, therefore we
+/// must use the `allow(dead_code)` decorator for this struct.
```
vaporos, line 73: This comment should probably be "//" not "///" as it isn't intended to be API doc.

sdk/src/workflow/state.rs
```diff
@@ -15,6 +15,9 @@
 #[derive(Clone)]
 pub struct WorkflowState {
     name: String,
+    #[allow(dead_code)]
```
vaporos, line 18: This could probably be removed instead, since its not used in v0.2.

shannynalayna: Removed this field in the next commit

cli/src/actions/products.rs
shannynalayna: Updated this comment to use "//" instead

dnewh (approved):

agunde406 (approved):

PR #1186, opened 2021-12-15
Fix CLI display of many results

dnewh:
This fixes the performance of the CLI when attempting to display many purchase order results. Previously, no results would be shown until all records were fetched from storage. With this update, a special iterator is used to take advantage of the REST API's paging and cache records so the CLI can print as soon as the first result comes back. This required significant rewrites to the CLI printing functionality.


CLEckhardt (approved):
Tested and works for me.

shannynalayna (approved):
Tested on sawtooth, looks good

agunde406 (approved):

PR #1199, opened 2021-12-22
Purchase Order protobuf updates

shannynalayna:
This change  makes updates to the "CreatePurchaseOrderPayload" and "UpdatePurchaseOrderPayload". The "create" payload has a new field, `workflow_type`, which allows a user to select which workflow the purchase order should be in. The "update" payload also has a new field, `version_updates`. This new field is a list of `UpdateVersionPayload`s to allow a user to update multiple versions of a purchase order within a single transaction.

### DIFF
--- a/cli/src/actions/purchase_order.rs
+++ b/cli/src/actions/purchase_order.rs
@@ -802,6 +802,7 @@ pub mod tests {
             .with_seller_org_id("seller".to_string())
             .with_workflow_state("created".to_string())
             .with_created_at(100)
+            .with_workflow_id("workflow".to_string())
             .build()
             .expect("Could not build create po payload");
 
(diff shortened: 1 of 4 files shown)


CLEckhardt (dismissed):
Tested and worked for me.

leebradley (dismissed):
LGTM

dnewh (dismissed):

cli/src/actions/purchase_order.rs
```diff
@@ -754,6 +754,7 @@ pub mod tests {
             .with_seller_org_id("seller".to_string())
             .with_workflow_state("created".to_string())
             .with_created_at(100)
+            .with_workflow_type("workflow".to_string())
```
vaporos, line 757: Using "workflow_id" might be more consistent with our use of _id throughout the rest of Grid.

shannynalayna: Updated this to be `workflow_id` and added a story to rename this field in the purchase order object, which is currently "workflow_type"

leebradley (approved):

vaporos (approved):

dnewh (approved):

PR #1213, opened 2021-12-30
Update "update po" action in po smart contract

shannynalayna:
Updates the "update_purchase_order" method to handle a list of versions. This also updates the method to correctly handle versions depending on the intended update state. This also updates the related tests.

### DIFF
--- a/contracts/purchase_order/src/handler.rs
+++ b/contracts/purchase_order/src/handler.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashMap;
+
 cfg_if! {
     if #[cfg(target_arch = "wasm32")] {
         use sabre_sdk::ApplyError;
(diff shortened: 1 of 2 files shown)


dnewh (approved):
tested

CLEckhardt (dismissed):
Tested

contracts/purchase_order/src/handler.rs
```diff
@@ -463,120 +556,317 @@ fn update_purchase_order(
...
+                if vers.is_draft() {
+                    return Err(ApplyError::InvalidTransaction(format!(
+                        "Purchase order {} version {} is a draft, \
+                        cannot accept a draft version",
+                        purchase_order.uid(),
+                        vers.version_id(),
+                    )));
+                }
+                vers.into_builder()
+                    .with_workflow_state("accepted".to_string())
```
vaporos, line 849: The smart contract should not know about specific workflow state strings like "accepted".

vaporos (changes requested):

contracts/purchase_order/src/handler.rs
shannynalayna: Updated to return errors if the smart contract ends up in a place where it finds other versions (besides the one specified in the po's accepted vers num) that are in a state with the `accepted` constraint or if the po's specified accepted version is not in a state with the `accepted` constraint

CLEckhardt (approved):
Tested

dnewh (approved):

vaporos (approved):

PR #1251, opened 2022-01-18
Add `StartWorkflowState`

shannynalayna:
This adds a new struct to Workflow state, `StartWorkflowState`. This is specifically used to validate a user's permissions to create an object within a workflow.

Also updates the methods within the permission checker to better specify between checking a user's permission within a workflow and when entering a workflow. This updates the `check_permission_with_workflow` method to be `check_permission_within_workflow` and adds a new method, `check_permission_to_enter_workflow` which inspects a `StartWorkflowState` rather than a `WorkflowState`. 

This also updates the purchase order smart contract handler to use the correct permission checking method (the create_X actions now use the `check_permission_to_enter_workflow` method) and updates an affected test that was specifying the "create" workflow state which no longer exists in the built-in subworkflows as it was replaced by the 'start_state'

### DIFF
--- a/sdk/src/pike/permissions/mod.rs
+++ b/sdk/src/pike/permissions/mod.rs
@@ -17,7 +17,7 @@ use crate::pike::addressing::{compute_agent_address, compute_role_address};
 use crate::pike::permissions::error::PermissionCheckerError;
 use crate::protocol::pike::state::{Agent, AgentList, Role, RoleList};
 use crate::protos::FromBytes;
-use crate::workflow::WorkflowState;
+use crate::workflow::{StartWorkflowState, WorkflowState};
 
 cfg_if! {
     if #[cfg(target_arch = "wasm32")] {
@@ -196,7 +196,7 @@ impl<'a> PermissionChecker<'a> {
         }
     }
 
-    pub fn check_permission_with_workflow(
+    pub fn check_permission_within_workflow(
         &self,
         permission: &str,
         signer: &str,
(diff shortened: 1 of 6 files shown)


sdk/src/pike/permissions/mod.rs
```diff
@@ -1043,9 +1043,8 @@ mod tests {
             buyer.add_transition("issued");
 
             let mut seller = PermissionAlias::new("po::seller");
-            seller.add_permission("can-create-po-version");
-            buyer.add_permission("can-transition-issued");
-            buyer.add_transition("issued");
+            seller.add_permission("can-transition-issued");
```
CLEckhardt, line 1046: Are these seller permissions needed?

CLEckhardt (approved):
Tested, looks good

sdk/src/pike/permissions/mod.rs
shannynalayna: Ah I don't think they are necessarily needed for the test, if that's what you mean. I removed the 'can-create-po-version' after we discussed that perm didn't make sense for a seller and had intended to remove it from the po built-in workflow as well. And the transition stuff was being misassigned to the 'buyer' perm alias which was fixed in the smart contract but not in this test. Does that answer your question?

CLEckhardt: Yes, thanks

leebradley (approved):

dnewh (approved):

PR #1252, opened 2022-01-19
Add `download-xsd` CLI command

leebradley:
Add an XSD download CLI command, along with several utility methods for
downloading, extracting, and validating the files within the schema
archive.

### DIFF
--- a/cli/man/grid-download-xsd.1.md
+++ b/cli/man/grid-download-xsd.1.md
@@ -1,0 +1,79 @@
+% GRID-DOWNLOAD-XSD(1) Cargill, Incorporated | Grid
+
+<!--
+  Copyright 2022 Cargill Incorporated
+  Licensed under Creative Commons Attribution 4.0 International License
+  https://creativecommons.org/licenses/by/4.0/
+-->
+
+NAME
+====
+
+**grid-download-xsd** - Downloads and extracts the XSDs necessary for Grid
+validation.
+
+SYNOPSIS
+========
+
+**grid download-xsd** \[**FLAGS**\] \[**OPTIONS**\] 
+
+DESCRIPTION
+===========
+
+This command downloads GS1 XSD files used by various Grid features. The
+downloaded artifacts are first copied into a cache directory. They are then
+expanded into Grid's state directory. If the desired artifacts are in the
+cache directory, Grid will not attempt to re-download them, and instead
+prefer the cache contents.
+
+To avoid downloading from the internet (for example, if a firewall rule
+would prevent access to the remote website), use the --copy-from and
+--no-download arguments.
+
+If --copy-from is used without --no-download, artfacts will be copied from
+the directory provided via --copy-from and any missing artifacts will be
+downloaded as usual.
+
+FLAGS
+=====
+
+`--no-download`
+: Do not download the XSD even if there is no artifact cached
+
+`--force`
+: Continue even if a checksum on the cached file is incorrect
+
+`-h`, `--help`
+: Prints help information.
+
+`-q`, `--quiet`
+: Do not display output.
+
+`-V`, `--version`
+: Prints version information.
+
+`-v`
+: Log verbosely.
+
+OPTIONS
+=======
+
+`--copy-from`
+: Replenish the cache from a directory resource and use that. The directory
+  should contain the following files:
+  /GS1\_XML\_3-4-1\_Publication.zip
+
+ENVIRONMENT VARIABLES
+=====================
+
+**`GRID_CACHE_DIR`**
+: Specifies the local path to the directory containing GRID cache.
+  The default value is "/var/cache/grid".
+
+**`GRID_STATE_DIR`**
+: Specifies the local path to the directory containing GRID state.
+  The default value is "/var/lib/grid".
+
+SEE ALSO
+========
+| Grid documentation: https://grid.hyperledger.org/docs/0.3/
--- a/cli/src/actions/xsd_downloader/downloader.rs
+++ b/cli/src/actions/xsd_downloader/downloader.rs
@@ -1,0 +1,464 @@
+/*
+ * Copyright 2022 Cargill Incorporated
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -----------------------------------------------------------------------------
+ */
+use std::fs::{self, File};
+use std::io::{self, Cursor};
+use std::path::PathBuf;
+
+use reqwest::{blocking::Client, Url};
+
+use crate::error::CliError;
+
+/// Uses the Reqwest library to download a file
+pub fn download(url: &Url, file_name: &str) -> Result<(), CliError> {
+    let client = Client::new();
+    let res = client
+        .get(url.clone())
+        .send()
+        .map_err(|err| CliError::InternalError(err.to_string()))?;
+    if res.status().is_server_error() {
+        return Err(CliError::ActionError(format!(
+            "received server error {:?}",
+            res.status()
+        )));
+    }
+    let body = res
+        .bytes()
+        .map_err(|err| CliError::InternalError(err.to_string()))?;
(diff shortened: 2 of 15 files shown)


dnewh (dismissed):
tested

cli/src/actions/xsd_downloader/validator.rs
```diff
@@ -0,0 +1,77 @@
...
+use crate::error::CliError;
+
+/// Validate the hash of a file an return a useful error message if the hash is invalid
+///
+/// * `file` - The file to check the hash of
+/// * `hash` - The expected sha256 hash
+pub fn validate_hash(file: &Path, hash: &[u8; 32]) -> Result<(), CliError> {
+    info!("validating hash of {file}", file = file.to_string_lossy());
+    let mut hasher = Sha256::new();
+    let data = std::fs::read(file).unwrap();
```
CLEckhardt, line 15: --no-download panics when there isn't an artifact

```
root@gridd-alpha:/usr/share/grid/xsd# grid download-xsd --no-downl│
oad                                                               │
validating hash of /var/lib/grid/artifact/gs1.zip                 │
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` v│
alue: Os { code: 2, kind: NotFound, message: "No such file or dire│
ctory" }', cli/src/actions/xsd_downloader/validator.rs:15:36      │
note: run with `RUST_BACKTRACE=1` environment variable to display │
a backtrace      
```

cli/src/main.rs
```diff
@@ -1775,6 +1777,50 @@ fn run() -> Result<(), CliError> {
...
+                        .conflicts_with("copy_from")
+                        .help("Never download a file, only extract the cached file"),
+                )
+                .arg(
+                    Arg::with_name("no_checksum")
+                        .long("no-checksum")
+                        .help("Do not perform a checksum on the cached file"),
+                )
+                .arg(
+                    Arg::with_name("copy_from")
```
CLEckhardt, line 1810: It looks like this should take a value.

cli/src/main.rs
```diff
@@ -1775,6 +1777,50 @@ fn run() -> Result<(), CliError> {
...
+                        .conflicts_with("force_download")
+                        .conflicts_with("no_download")
+                        .long("copy-from")
+                        .help("Replenish the cache from a directory resource and use that"),
+                )
+                .arg(
+                    Arg::with_name("artifact_dir")
```
CLEckhardt, line 1817: It looks like this should also take a value.

CLEckhardt (changes requested):

cli/src/actions/xsd_downloader/validator.rs
leebradley: Good catch! Fixed

cli/src/main.rs
leebradley: Added `takes_value(true)`

cli/src/main.rs
leebradley: Added `takes_value(true)`

cli/man/grid-download-xsd.1.md
```diff
@@ -0,0 +1,77 @@
+% GRID-KEYGEN(1) Cargill, Incorporated | Grid
+
+<!--
+  Copyright 2021 Cargill Incorporated
```
shannynalayna, line 4: This man page should be copyright for `2022`

cli/src/actions/xsd_downloader/downloader.rs
```diff
@@ -0,0 +1,419 @@
+use std::fs::{self, File};
```
shannynalayna, line 1: This file is missing a copyright

cli/src/actions/xsd_downloader/extractor.rs
```diff
@@ -0,0 +1,264 @@
+use std::fs::{self, File};
```
shannynalayna, line 1: Need a copyright

cli/src/actions/xsd_downloader/mod.rs
```diff
@@ -0,0 +1,603 @@
+mod downloader;
```
shannynalayna, line 1: Copyright :)

shannynalayna: Copyrights need to be added to all the new files, noticed a few more but won't comment everywhere

cli/src/main.rs
```diff
@@ -1775,6 +1774,54 @@ fn run() -> Result<(), CliError> {
...
+                .arg(
+                    Arg::with_name("force")
+                        .long("force")
+                        .help("Continue even if a checksum on the cached file is incorrect"),
+                )
+                .arg(
+                    Arg::with_name("copy_from")
+                        .takes_value(true)
+                        .conflicts_with("force_download")
+                        .conflicts_with("no_download")
```
shannynalayna, line 1800: From the man page, it looks like this option should be usable with `no_download`. The man page should be updated to show these args are not usable together, unless this line is a mistake?

cli/man/grid-download-xsd.1.md
leebradley: Updated!

cli/src/actions/xsd_downloader/downloader.rs
leebradley: Added!

cli/src/actions/xsd_downloader/extractor.rs
leebradley: Added!

cli/src/actions/xsd_downloader/mod.rs
leebradley: Added!

cli/src/main.rs
leebradley: So actually I'd somehow missed the req that no_download should be compatible with copy_from, so you found a legit bug here. Fixed and added a unit test to validate the functionality.

shannynalayna (dismissed):
Tested

CLEckhardt (commented):
Tested and functionality is good. A few notes:
- `--copy-from` without `--no-download` effectively ignores the arg if the filename is incorrect
- `--force` still gives a message “validating hash of ...”
- man page and CLI help should probably say that the file for `--copy-from` needs to be named “gs1.zip”
Still working on testing po actions, running into some unrelated issues, but I'll confirm it all works when that's done

leebradley:
> * `--copy-from` without `--no-download` effectively ignores the arg if the filename is incorrect

Yeah, I was initially going to be very strict and require all files exist, but there's also the use case @shannynalayna pointed out of only having some of the files available (and downloading the rest). I'm not sure how to best approach this.

> * `--force` still gives a message “validating hash of ...”

By design `--force` still validates the hash of downloaded files, and only ignores the hashes of cached files.

> * man page and CLI help should probably say that the file for `--copy-from` needs to be named “gs1.zip”

Hmm if this is going to be added to docs that means we'll technically be supporting manually tinkering with the internal cache, which tbh I didn't anticipate. I think I'm going to rename this file so it's a little more future-proof.

CLEckhardt:
Cool, makes sense.

leebradley:
So in response to the feedback form @CLEckhardt I've made three changes:

1) After consulting a bit, we've decided to support users working with these cache files. Thus I've renamed "gs1.zip" in the cache to "GS1_XML_3-4-1_Publication.zip" so it matches the default filename and will not need renamed in the future if a different version comes out.

2) Added a reference to GS1_XML_3-4-1_Publication.zip in the docs so users are aware what needs to be in the copy directory.

3) Changed a `debug!` message to `info!` so users can immediately see if there are missing files in the cache directory. Example below:

```
root@gridd-alpha:/# grid download-xsd --copy-from foo              
file "GS1_XML_3-4-1_Publication.zip" doesn't exist in copy path
downloading file from https://www.gs1.org/docs/EDI/xml/3.4.1/GS1_XML_3-4-1_Publication.zip
download finished
validating hash of /var/cache/grid/xsd_artifact_cache/GS1_XML_3-4-1_Publication.zip.download
extracting to schema directory
```

cli/man/grid-download-xsd.1.md
```diff
@@ -0,0 +1,79 @@
...
+
+**grid download-xsd** \[**FLAGS**\] \[**OPTIONS**\] 
+
+DESCRIPTION
+===========
+
+This command downloads GS1 XSD files used by various Grid features. The
+downloaded artifacts are first copied into a cache directory. They are then
+expanded into Grid's state directory. If the desired artifacts are in the
+cahce directory, Grid will not attempt to re-download them, and instead
```
shannynalayna, line 26: sp/cahce/cache

leebradley: Fixed!

shannynalayna (approved):
Tested; temp file is being removed if the overall download fails

dnewh (approved):

PR #1255, opened 2022-01-19
Fix PO migrations

dnewh:
This updates the PO migrations by adding a new migrations file for the
0.3 version of the tables and reverts the original file back to its
original 0.2 state.


shannynalayna (approved):
Tested; was able to upgrade from 0-2 and created a purchase order

leebradley (approved):
Diffed against 0-2 branch. LGTM

PR #1258, opened 2022-01-21
Update default schema dir to `/var/lib/grid/xsd`

leebradley:
The schema directory needs to be writable by Grid by default, which is
not necessarily the case with the previous directory of
`/usr/share/grid/xsd`. This change modifies the default to be
`/var/lib/grid/xsd`.


leebradley:
The changes in this PR will be necessary due to the removal of XSDs in https://github.com/hyperledger/grid/pull/1257 and the addition of the download utility in https://github.com/hyperledger/grid/pull/1252

leebradley:
Closing because these changes will be in #1252

Issue #1246, opened 2022-01-13
Fix `po update` removing accepted version


CLEckhardt:
In system of record, on the po update command, a buyer can remove the accepted version without changing the version state from 'accepted', resulting in an invalid state.

PR #1265, opened 2022-01-25
Add check if po doesn't have accepted version

shannynalayna:
This change adds a check if the purchase order does not have the
`accepted_version_number` set to validate the list of versions for that
purchase order do not contain versions that exist in an
`accepted`-constrained state.

This addresses this bug: https://github.com/hyperledger/grid/issues/1246


leebradley (approved):

leebradley:
Tested locally with https://github.com/hyperledger/grid/pull/1266

dnewh (approved):

PR #1266, opened 2022-01-26
Add test to validate removing accepted version

leebradley:
Validate that the po versions are consistently not accepted with the
parent po when the po is no longer accepted.


leebradley:
This test will fail until https://github.com/hyperledger/grid/issues/1246 is fixed

shannynalayna (approved):

dnewh (approved):

PR #1268, opened 2022-01-26
Check if version exists in create version command

shannynalayna:
This change updates how the revision ID is fetched for a version being
created. As this version is being created, the revision ID should be '1'
rather than the most recent revision ID of the version. The most recent
revision ID should not exist, as a version being created should not
already exist.

Testing instructions:
1. Create a PO & a PO version
2. Attempt to create the same version made previously
3. You should receive an error that that version already exists. Previously, you would receive the error I pasted in a comment below.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2982,14 +2982,20 @@ fn run() -> Result<(), CliError> {
                         )?;
                     }
 
-                    let workflow_state = m.value_of("workflow_state").unwrap();
-
-                    let revision_id = purchase_order::get_latest_revision_id(
-                        &*purchase_order_client,
-                        &po,
-                        version_id,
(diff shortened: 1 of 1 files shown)


shannynalayna:
I was running into an error when I attempted to create the same version twice. The error: 
```
grid_purchase_order received InvalidTransaction: Revision IDs begin incrementing at `1`, attempting to create a new version with revision ID `2`
```
What's happening, in the CLI command we were fetching the most recent revision ID for a version being created. This is invalid as the version does not already exist (we're creating it) and so the revision ID should be 1. The error added will now better represent that the version already exists.

dnewh (approved):
tested

leebradley (approved):

PR #1272, opened 2022-01-28
Remove unwraps for required arguments

leebradley:
Previously required arguments were retrieved via a call to unwrap(),
which had the capacity to panic at runtime. This change removes those
unwrap() calls and replaces them with a function that sends back a
standard CliError type.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1891,8 +1891,8 @@ fn run() -> Result<(), CliError> {
                 let wait = value_t!(m, "wait", u64).unwrap_or(0);
 
                 let create_agent = CreateAgentActionBuilder::new()
-                    .with_org_id(m.value_of("org_id").unwrap().into())
-                    .with_public_key(m.value_of("public_key").unwrap().into())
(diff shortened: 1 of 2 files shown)


peterschwarz (approved):

cli/src/main.rs
```diff
@@ -2975,7 +2979,7 @@ fn run() -> Result<(), CliError> {
                     validate_order_xml_3_4(&xml_str, false, &data_validation_dir)?;
                     info!("Purchase order was valid.");
 
-                    let version_id = m.value_of("version_id").unwrap();
+                    let version_id = value_of_required(m, "version_id")?;
 
                     let mut po = m.value_of("po").map(String::from).unwrap();
```
shannynalayna, line 2984: Aren't the PO UID args required for each of these subcommands? Should probably use the added fn for these ones too

leebradley: Fixed!

CLEckhardt (approved):

shannynalayna (approved):

dnewh (approved):

PR #1307, opened 2022-02-14
Replace node-sass@4.13.1 with sass@1.49.7

leebradley:
This change replaces node-sass due to a dependabot alert of
vulnerability CVE-2020-24025, where certificate validation was disabled
when requesting binaries.

Since node-sass is deprecated, it was replaced with sass.


leebradley:
Diffed the resulting CSS output and it is identical.

shannynalayna (approved):

dnewh (approved):

PR #1330, opened 2022-02-23
Add migrations for new batch tracking tables

dnewh:
This adds postgres and sqlite migrations for updated batch tracking
tables. These migrations remove the unused and outdated tables of the
same name and add updated ones.


leebradley (approved):
LGTM

PR #1345, opened 2022-03-15
Add `GET` proxy endpoint for Griddle 

shannynalayna:
Add ProxyClient trait and implementation, add endpoint for proxying GET requests, add this endpoint as the default service to the Griddle REST API for `GET` requests.


sdk/src/client/mod.rs
```diff
@@ -36,6 +36,10 @@ pub mod reqwest;
 pub mod schema;
 #[cfg(feature = "schema")]
 pub use schema::*;
+#[cfg(feature = "proxy-client")]
+pub mod proxy;
+#[cfg(feature = "proxy-client")]
+pub use proxy::*;
```
vaporos, line 42: Expand the wildcard imports.

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    fn merge_request_into_reqwest_builder(
+        &self,
+        client: &ReqwestBlockingClient,
+        p_req: ProxyRequest,
+    ) -> Result<ReqwestBuilder, ErrorResponse> {
+        // Generate the request URL
+        let mut url = ReqwestUrl::parse(&self.gridd_url)
```
vaporos, line 53: This should be created in ReqwestProxyClient::new() so the parse error occurs there, because its a configuration error (should probably be InvalidArgumentError).

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::iter::FromIterator;
+
+use reqwest::{
```
vaporos, line 17: IMO avoid renaming in use statements unless there is a conflict.

sdk/src/client/proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub query_params: Option<String>,
+    pub body: Option<ProxyRequestBody>,
+    pub method: ProxyMethod,
+}
+
+#[derive(Debug)]
+pub enum ProxyRequestBody {
+    Json(String),
+    Bytes(Vec<u8>),
+}
```
vaporos, line 45: This could probably be reworked such that if we add xml later (as an example), we don't need to change the request body enum/struct, by breaking this down into its more fundamental HTTP parts.

sdk/src/client/proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use serde_json::Value as JsonValue;
+
+use crate::client::Client;
+use crate::rest_api::resources::error::ErrorResponse;
+
+pub trait ProxyClient: Client + Send {
```
vaporos, line 20: Does this really have anything to do with Client?

sdk/src/rest_api/actix_web_3/run.rs
```diff
@@ -12,22 +12,39 @@
...
 
+#[cfg(feature = "proxy")]
+use crate::client::ProxyClient;
 use crate::error::InternalError;
 
+#[cfg(feature = "proxy")]
+use super::routes::proxy_get;
 use super::{routes::submit, KeyState, StoreState};
 
 pub async fn run(
```
vaporos, line 27: This run() function should probably be in griddle.

sdk/src/rest_api/actix_web_3/routes/proxy.rs
```diff
@@ -0,0 +1,90 @@
...
+impl TryFrom<&ActixMethod> for ProxyMethod {
+    type Error = ErrorResponse;
+
+    fn try_from(actix_method: &ActixMethod) -> Result<Self, Self::Error> {
+        match *actix_method {
+            ActixMethod::GET => Ok(ProxyMethod::GET),
+            ActixMethod::POST => Ok(ProxyMethod::POST),
+            ActixMethod::PUT => Ok(ProxyMethod::PUT),
+            ActixMethod::DELETE => Ok(ProxyMethod::DELETE),
+            _ => Err(ErrorResponse::new(400, "Invalid endpoint method provided")),
```
vaporos, line 87: Why not support the full set here?

vaporos: This is the only reason TryFrom is here and on the containing request instead of From.

sdk/src/rest_api/actix_web_3/routes/proxy.rs
```diff
@@ -0,0 +1,90 @@
...
+use actix_web::{http::Method as ActixMethod, http::StatusCode, web, HttpRequest, HttpResponse};
+
+use crate::client::{HeaderList, ProxyClient, ProxyMethod, ProxyRequest};
+use crate::rest_api::resources::error::ErrorResponse;
+
+pub async fn proxy_get(
+    req: HttpRequest,
+    proxy_client: web::Data<Box<dyn ProxyClient>>,
+) -> HttpResponse {
+    let request = match ProxyRequest::try_from(req) {
```
vaporos, line 26: This should be changed to ProxyRequest::from(req) so the Err branch becomes unnecessary.

sdk/src/rest_api/actix_web_3/routes/proxy.rs
```diff
@@ -0,0 +1,90 @@
...
+        Err(err) => {
+            return HttpResponse::build(
+                StatusCode::from_u16(err.status_code())
+                    .unwrap_or(StatusCode::INTERNAL_SERVER_ERROR),
+            )
+            .json(err);
+        }
+    };
+    debug!("{:?}", request);
+    match proxy_client.proxy(request) {
```
vaporos, line 37: Can the Err branch here only occur when unable to submit the request to the proxied server? If so, this should always be 503.

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl ProxyClient for ReqwestProxyClient {
+    fn proxy(&self, req: ProxyRequest) -> Result<JsonValue, ErrorResponse> {
+        let client = ReqwestBlockingClient::new();
+
+        let request = self.merge_request_into_reqwest_builder(&client, req)?;
+
+        let response = request
+            .send()
+            .map_err(|err| ErrorResponse::new(500, &format!("Request failed to send: {err}")))?;
```
vaporos, line 94: This should probably be 502.

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        if !response.status().is_success() {
+            let error_response: ErrorResponse = response.json().map_err(|err| {
+                ErrorResponse::new(500, &format!("Failed to get response data: {err}"))
+            })?;
+            return Err(error_response);
+        }
+
+        let json_response = response.json().map_err(|err| {
```
vaporos, line 103: A response object should be returned here, and the conversion to json avoided completely.

sdk/src/client/mod.rs
shannynalayna: Removed use of wildcard imports

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+use serde_json::Value as JsonValue;
+
+use crate::client::reqwest::post_batches;
+use crate::client::{Client, ProxyClient, ProxyMethod, ProxyRequest, ProxyRequestBody};
+use crate::error::ClientError;
+use crate::rest_api::resources::error::ErrorResponse;
+
+/// The Reqwest implementation of the Proxy client
+pub struct ReqwestProxyClient {
+    gridd_url: String,
```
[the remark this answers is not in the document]
shannynalayna: removed references to `gridd` url

sdk/src/client/reqwest/proxy/mod.rs
shannynalayna: Updated the ReqwestProxyClient::new fn to parse the url initially

sdk/src/client/reqwest/proxy/mod.rs
shannynalayna: Updated renaming structs where it did not conflict, although kept the `BlockingReqwestClient` rename as this clarifies the reqwest client being used and this matches the use/renaming in the use stmt where it is used elsewhere in Grid

sdk/src/client/proxy.rs
shannynalayna: I changed this around to hold a field for the `content_type` and the content itself. Thinking the content-type val would allow for the body to be parsed as needed,

sdk/src/client/proxy.rs
shannynalayna: Removed this trait from `ProxyClient` and removed the proxy impl out of the sdk/src/client directory

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+            reqwest_builder = reqwest_builder.body(ReqwestBody::from(body));
+        }
+
+        Ok(reqwest_builder)
+    }
+}
+
+impl ProxyClient for ReqwestProxyClient {
+    fn proxy(&self, req: ProxyRequest) -> Result<JsonValue, ErrorResponse> {
+        let client = ReqwestBlockingClient::new();
```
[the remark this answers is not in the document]
shannynalayna: I did not run into any lifetime thing moving creating the client into the try_from impl. The client is stored in that request builder so the proxy fn itself doesn't need to also create a client, which is neat

sdk/src/rest_api/actix_web_3/run.rs
shannynalayna: Before we move this to Griddle, perhaps we could discuss how we want the rest_api in Griddle to be set up? This run fn is implemented with actix-web-3, so detangling it from this dependency (as we plan to impl actix 4, etc.) will require a little more work around getting the fn to start griddle + its rest api without entirely depending on actix web 3. Unless, we just want to port over this run fn into a rest_api/actix_web_3 module to use for the time being

sdk/src/rest_api/actix_web_3/routes/proxy.rs
shannynalayna: Add a `Custom(Vec<u8>)` variant here plus the other main methods, besides the ones already included. I did end up having to implement this conversion as Tryrom because some (2, the Extensions) variants are private structs. Therefore, I had to create those private structs with the pub `from_bytes` method. This fn returns a result, so I could not impl this just as From

sdk/src/rest_api/actix_web_3/routes/proxy.rs
shannynalayna: Was able to update this to be a From impl for ProxyRequest

sdk/src/client/reqwest/proxy/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        if let Some(body) = p_req.body {
+            reqwest_builder = reqwest_builder.body(ReqwestBody::from(body));
+        }
+
+        Ok(reqwest_builder)
+    }
+}
+
+impl ProxyClient for ReqwestProxyClient {
+    fn proxy(&self, req: ProxyRequest) -> Result<JsonValue, ErrorResponse> {
```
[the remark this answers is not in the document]
shannynalayna: Ended up creating a struct, `ProxyResponse` that holds a status code and message that can be deserialized by the endpoint. This also allows for this fn to just return ProxyResponse rather than a result

sdk/src/client/reqwest/proxy/mod.rs
shannynalayna: Removed this conversion from the reqwest proxy client

CLEckhardt (dismissed):

sdk/src/rest_api/actix_web_3/routes/proxy.rs
shannynalayna: The reqwest client may fail to send the request for multiple reasons, perhaps this function should parse them a bit more because right now if the send method runs into an error, we return a 500. But, the proxy function may also return a 400, as it builds the request. There are multiple `Kind`s for the reqwest error: https://docs.rs/reqwest/latest/src/reqwest/error.rs.html#232

sdk/src/client/reqwest/proxy/mod.rs
shannynalayna: I kind of responded to this above, but the send function may return errors for all of these reasons: https://docs.rs/reqwest/latest/src/reqwest/error.rs.html#232 (Linking the kinds of errors reqwest may return). Again, perhaps the error from the reqwest send function should be parsed more thoroughly to accurately reflect the err encountered

vaporos (approved):

dnewh (approved):

PR #1349, opened 2022-03-22
Add `get_batch` db operation

dnewh:
This adds an implementation for the `BatchTrackingStore` `get_batch()` operation. This also adds a unit test to test adding and fetching a batch.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/operations/get_batch.rs
+++ b/sdk/src/batch_tracking/store/diesel/operations/get_batch.rs
@@ -1,0 +1,321 @@
+// Copyright 2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


sdk/src/batch_tracking/store/diesel/operations/get_batch.rs
```diff
@@ -0,0 +1,297 @@
...
+{
+    fn get_batch(
+        &self,
+        id: &str,
+        service_id: &str,
+    ) -> Result<Option<TrackingBatch>, BatchTrackingStoreError> {
+        self.conn.transaction::<_, BatchTrackingStoreError, _>(|| {
+            let query = batches::table
+                .into_boxed()
+                .left_join(
```
shannynalayna, line 51: Since the queries in this op are pretty length and coming from multiple tables, could you maybe add some comments to describe what is being loaded /selected /filtered etc.

dnewh: added

shannynalayna (approved):

leebradley (approved):

PR #1351, opened 2022-03-25
Add `get_batch_status` db operation

dnewh:
This implements the `get_batch_status` operation for the Batch Tracking
    store. This operation returns the current batch status for a batch with
    a given ID and service ID. This PR also contains several other small but
    necessary updates to the models to make this work.
    
### DIFF
--- a/sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
+++ b/sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
@@ -1,0 +1,226 @@
+// Copyright 2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::BatchTrackingStoreOperations;
+use crate::error::InternalError;
+
+use crate::batch_tracking::store::diesel::{
+    models::{BatchStatusModel, TransactionReceiptModel},
+    schema::{batch_statuses, transaction_receipts, transactions},
+    BatchStatus, InvalidTransaction, TransactionReceipt, ValidTransaction,
+};
+
+use crate::batch_tracking::store::BatchTrackingStoreError;
+use diesel::prelude::*;
+use std::convert::TryFrom;
+
+pub(in crate::batch_tracking::store::diesel) trait BatchTrackingStoreGetBatchStatusOperation {
+    fn get_batch_status(
+        &self,
+        id: &str,
+        service_id: &str,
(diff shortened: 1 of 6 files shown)


sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
```diff
@@ -0,0 +1,210 @@
...
+                )
+                .select((
+                    batch_statuses::all_columns,
+                    transactions::all_columns.nullable(),
+                    transaction_receipts::all_columns.nullable(),
+                ));
+
+            let query_result: Vec<(
+                BatchStatusModel,
+                Option<TransactionModel>,
```
shannynalayna, line 70: It doesn't look like we're using the txn model that's being returned here. Why is that being returned with these objects?

dnewh: These are used to join transaction receipts with the batch

sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
```diff
@@ -0,0 +1,218 @@
...
+
+            // This query fetches the transactions and any associated receipts
+            // for the given batch ID and service ID
+            let txn_query = transactions::table
+                .into_boxed()
+                .left_join(
+                    transaction_receipts::table
+                        .on(transaction_receipts::transaction_id.eq(transactions::transaction_id)),
+                )
+                .select((
```
shannynalayna, line 78: Could we update this select statement to not include the transactions if we aren't using them at all? Seems like it would be a little faster if diesel doesn't have to load all of the txns in this query

shannynalayna: It also looks like we aren't matching on the service ID in this query

sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
```diff
@@ -0,0 +1,218 @@
...
+                    )))
+                })?;
+
+            if batch_status_model.is_none() {
+                return Ok(None);
+            }
+
+            // This query fetches the transactions and any associated receipts
+            // for the given batch ID and service ID
+            let txn_query = transactions::table
```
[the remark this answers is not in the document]
dnewh: yep, accidentally missed the filter. Added

sdk/src/batch_tracking/store/diesel/operations/get_batch_status.rs
dnewh: This query needs to do a join with the transactions table b/c the receipts table does not have batch_id as a fk

sdk/src/batch_tracking/store/mod.rs
```diff
@@ -39,28 +39,40 @@ pub enum BatchStatus {
 #[derive(Clone, Debug, PartialEq)]
 pub struct InvalidTransaction {
     transaction_id: String,
-    error_message: String,
-    error_data: Vec<u8>,
+    error_message: Option<String>,
+    error_data: Option<Vec<u8>>,
```
shannynalayna, line 43: Can you add a comment here to explain the difference between these two errors?

dnewh: added

shannynalayna (approved):

leebradley (approved):

PR #1374, opened 2022-04-05
Add `change_batch_to_submitted` operation

dnewh:
This implements the `change_batch_to_submitted` DB operation for the
    Batch Tracking store. This includes a couple unit tests for the
    operation as well.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/operations/change_batch_to_submitted.rs
+++ b/sdk/src/batch_tracking/store/diesel/operations/change_batch_to_submitted.rs
@@ -1,0 +1,349 @@
+// Copyright 2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::BatchTrackingStoreOperations;
+use crate::error::InternalError;
+
+use crate::batch_tracking::store::{
+    diesel::{
+        models::{
+            NewBatchStatusModel, NewSubmissionModel, TransactionModel, TransactionReceiptModel,
+        },
+        schema::{batch_statuses, batches, submissions, transaction_receipts, transactions},
+    },
+    BatchStatus, BatchStatusName, BatchTrackingStoreError,
+};
+use diesel::{
+    dsl::{exists, insert_into, update},
+    prelude::*,
+    select,
+};
+
+pub(in crate::batch_tracking::store::diesel) trait BatchTrackingStoreChangeBatchToSubmittedOperation
+{
+    fn change_batch_to_submitted(
+        &self,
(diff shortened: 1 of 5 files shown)


CLEckhardt (approved):

sdk/src/batch_tracking/store/diesel/operations/change_batch_to_submitted.rs
```diff
@@ -0,0 +1,353 @@
...
+        status: Option<NewBatchStatusModel>,
+        submission: NewSubmissionModel,
+    ) -> Result<(), BatchTrackingStoreError> {
+        self.conn.transaction::<_, BatchTrackingStoreError, _>(|| {
+            let submitted_statuses = vec![
+                BatchStatus::Pending.to_string(),
+                BatchStatus::Invalid(Vec::new()).to_string(),
+                BatchStatus::Valid(Vec::new()).to_string(),
+                BatchStatus::Committed(Vec::new()).to_string(),
+            ];
```
leebradley, line 63: So this is awkward because the vector is being constructed for every function call, `.contains()` is being used instead of `match`, and the construction of enum components like `BatchStatus::Valid(Vec::new()).to_string()` to just extract a string representation.

Rather than do this you should probably make a new enum, like this example:

```rust
enum BatchStatusName {
    Unknown,
    Pending,
    Delayed,
    Invalid,
    Valid,
    Committed,
}

impl BatchStatusName {
    fn from_string(value: &str) -> Option<BatchStatusName> {
        match value {
            "Unknown" => Some(BatchStatusName::Unknown),
            "Pending" => Some(BatchStatusName::Pending),
            "Delayed" => Some(BatchStatusName::Delayed),
            "Invalid" => Some(BatchStatusName::Invalid),
            "Valid" => Some(BatchStatusName::Valid),
            "Committed" => Some(BatchStatusName::Committed),
            _ => None
        }
    }
}

fn main() {
    let status_string = "Invalid";
    let status = BatchStatusName::from_string(status_string).unwrap(); // Throw an error don't unwrap, this is just an example
    let submitted = match status {
        BatchStatusName::Pending |
        BatchStatusName::Invalid |
        BatchStatusName::Valid |
        BatchStatusName::Committed => true,
        _ => false,
    };
    
    println!("is it submitted? {:?}", submitted);
}
```

If you wanted to ensure this new enum perfectly matched the `BatchStatus` enum, you could write a unit test with a `match` statement between `BatchStatus` and `BatchStatusName` where rust would throw an error if they ever got out of sync (because all patterns aren't covered).

sdk/src/batch_tracking/store/diesel/operations/change_batch_to_submitted.rs
```diff
@@ -0,0 +1,353 @@
...
+
+            update(batches::table)
+                .filter(
+                    batches::batch_id
+                        .eq(&batch_id)
+                        .and(batches::service_id.eq(&service_id)),
+                )
+                .set(batches::submitted.eq(true))
+                .execute(self.conn)
+                .map(|_| ())
```
leebradley, line 347: At a glance this doesn't look necessary.

sdk/src/batch_tracking/store/diesel/operations/change_batch_to_submitted.rs
```diff
@@ -0,0 +1,353 @@
...
```
leebradley, line 192: At a glance this doesn't look necessary.

shannynalayna (approved):

leebradley (approved):

shannynalayna (approved):

PR #1380, opened 2022-04-08
Update batch tracking migrations

dnewh:
This updates the database schema for the batch tracking tables by
adding batch/transaction ID columns. These columns are intended to be
the primary keys for a table and consist of a hash of the batch/txn
header and the service ID. Columns for the service ID and batch/txn
headers are also included for deserialization sake. This is a necessary
change because diesel does not fully support composite keys.
Specifically, it does not implement the `BelongsToDsl` trait for
composite keys and they do not intend to implement this functionality,
as stated in https://github.com/diesel-rs/diesel/issues/1413.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/models.rs
+++ b/sdk/src/batch_tracking/store/diesel/models.rs
@@ -13,10 +13,11 @@
 // limitations under the License.
 
 use core::convert::TryFrom;
+use crypto::digest::Digest;
+use crypto::sha2::Sha512;
 
 use crate::batch_tracking::store::diesel::schema::*;
 use crate::error::InternalError;
-use chrono::NaiveDateTime;
 
 use super::{
     BatchStatus, InvalidTransaction, SubmissionError, TrackingBatch, TrackingTransaction,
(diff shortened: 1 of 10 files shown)


sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -381,3 +409,11 @@ pub fn make_transaction_models(batches: &[TrackingBatch]) -> Vec<TransactionMode
 
     models
 }
+
+pub fn make_database_id(id: &str, service_id: &str) -> String {
```
shannynalayna, line 413: Would this still enable searching for a batch via the data change ID? If someone were to try to fetch the batch using the data change ID over the batch ID

shannynalayna: Perhaps we could add a few sqlite in memory tests to ensure we'd be able to search for the batch using the data change id or the batch id

sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -125,7 +135,7 @@ impl From<TransactionModel> for TrackingTransaction {
         Self {
             family_name: transaction.family_name.to_string(),
             family_version: transaction.family_version.to_string(),
-            transaction_id: transaction.transaction_id.to_string(),
+            transaction_id: transaction.transaction_header.to_string(),
```
CLEckhardt, line 138: Should `TrackingTransaction.transaction_id` be `transaction_header` so they are consistent?

sdk/src/batch_tracking/store/diesel/models.rs
dnewh: this will not affect that at all. The data change ID field is still there so filtering on that is still available

sdk/src/batch_tracking/store/diesel/models.rs
dnewh: good call. updated

sdk/src/batch_tracking/store/diesel/models.rs
shannynalayna: If someone were to add a batch with batch ID `ABCDE` and service ID `XYZ`, that hash would be different if the batch had a data change ID of `012345` or whatever. It seems like if we're taking in the ID directly and hashing it, and if it doesn't match the original ID the batch was hashed with then it won't be returned. Think it would still be worthwhile to make some simple tests to validate that assumption.

leebradley (approved):

shannynalayna (approved):

PR #1381, opened 2022-04-11
Add `get_unsubmitted_batches` operation

dnewh:
This implements the `get_unsubmitted_batches()` operation for the
    batch tracking store. This operation gets all batches that have not yet
    been sumbitted. This is used by the batch queuer.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/models.rs
+++ b/sdk/src/batch_tracking/store/diesel/models.rs
@@ -18,12 +18,12 @@ use crate::batch_tracking::store::diesel::schema::*;
 use crate::error::InternalError;
 
 use super::{
-    BatchStatus, InvalidTransaction, SubmissionError, TrackingBatch, TrackingTransaction,
-    TransactionReceipt, ValidTransaction,
+    BatchStatus, InvalidTransaction, SubmissionError, TrackingBatch, TrackingBatchList,
+    TrackingTransaction, TransactionReceipt, ValidTransaction,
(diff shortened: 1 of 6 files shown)


sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -402,6 +404,121 @@ impl TryFrom<SubmissionModel> for SubmissionError {
...
+                } else {
+                    None
+                }
+            } else {
+                None
+            };
+
+            let txns: Vec<&TransactionModel> = transactions
+                .iter()
+                .filter(|t| t.batch_id == batch.batch_id && t.service_id == batch.service_id)
```
shannynalayna, line 478: Curious if this filtering should be done in the query ?

sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -402,6 +404,121 @@ impl TryFrom<SubmissionModel> for SubmissionError {
     }
 }
 
+impl TryFrom<&SubmissionModel> for SubmissionError {
```
shannynalayna, line 407: Could probably remove the above impl and just ref the model being converted that uses ^ the above impl

dnewh: good call. Updated

leebradley (approved):

shannynalayna (approved):

PR #1393, opened 2022-04-26
Add `get_failed_batches` operation

dnewh:
This adds an operation to get failed batches. This also adds a unit test for the operation.


shannynalayna (approved):
My only comment may be to ask if we could possibly figure out a way to incorporate the `failed_statuses` into the sql_query we're building. Would make updating the op simpler in the future but may just be a consequence of using the raw sql queries

leebradley (approved):

PR #1398, opened 2022-04-27
Add `clean_stale_records` operation

dnewh:
This adds an operation to clean batch tracking records that were created
    before a specified time. This works by removing batch records with a
    `created_at` value less than the specified timestamp. As the other
    batch tracking tables all use foreign keys derived from the batch_id and
    that use cascading deletes, this will remove the batch record and any
    associated "child" records.

### DIFF
--- a/sdk/src/batch_tracking/store/mod.rs
+++ b/sdk/src/batch_tracking/store/mod.rs
@@ -478,7 +478,7 @@ impl TrackingBatchBuilder {
             ));
         };
 
-        if created_at <= 0 {
+        if created_at < 0 {
             return Err(BatchBuilderError::MissingRequiredField(
                 "created_at".to_string(),
             ));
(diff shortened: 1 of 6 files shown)


sdk/src/batch_tracking/store/mod.rs
```diff
@@ -478,7 +478,7 @@ impl TrackingBatchBuilder {
             ));
         };
 
-        if created_at <= 0 {
+        if created_at < 0 {
```
shannynalayna, line 481: Why are we removing this constraint ?

leebradley (approved):

shannynalayna (dismissed):

shannynalayna (approved):

CLEckhardt (approved):

leebradley (approved):

PR #1400, opened 2022-04-29
Clean up `BatchTrackingStore` tests

dnewh:
This cleans up the batch tracking store tests by putting repetitive,
    shared code into functions.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/mod.rs
+++ b/sdk/src/batch_tracking/store/diesel/mod.rs
@@ -622,8 +622,8 @@ mod tests {
     use diesel::r2d2::{ConnectionManager, Pool};
     use diesel::sqlite::SqliteConnection;
     use transact::protocol::{
-        batch::BatchBuilder,
-        transaction::{HashMethod, TransactionBuilder},
+        batch::{Batch, BatchBuilder},
+        transaction::{HashMethod, Transaction, TransactionBuilder},
     };
 
     use crate::batch_tracking::store::{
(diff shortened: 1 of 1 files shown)


sdk/src/batch_tracking/store/diesel/mod.rs
```diff
@@ -1519,4 +1328,65 @@ mod tests {
...
+    }
+
+    fn get_transact_batch(signer: &dyn Signer, transactions: Vec<Transaction>) -> Batch {
+        BatchBuilder::new()
+            .with_transactions(transactions)
+            .build(&*signer)
+            .unwrap()
+    }
+
+    fn get_tracking_batch(
```
leebradley, line 1360: This function is the only weird one to me, just because the builder pattern is (in part) meant to avoid large number of optional arguments, and we already have a builder.

So right now the function signature is:
```rust
    fn get_tracking_batch(
        batch: Batch,
        submitted: bool,
        timestamp: Option<i64>,
        batch_status: Option<BatchStatus>,
        submission_error: Option<SubmissionError>,
        dcid: Option<String>,
    ) -> TrackingBatch { /* snip */ }
```

Maybe it would be cleaner to do something like this, where you'd just preload the builder with test and required values:

```rust
    fn get_tracking_batch_builder(
        batch: Batch,
        submitted: bool,
    ) -> TrackingBatchBuilder { /* snip */ }
```

leebradley (changes requested):

sdk/src/batch_tracking/store/diesel/mod.rs
dnewh: updated

leebradley (approved):
LGTM

shannynalayna (changes requested):
Should update all the `unwrap`s that happen throughout the tests to be `expect`s for easier debugging

shannynalayna (approved):

PR #1401, opened 2022-04-29
Make `TrackingBatch` `service_id` an `Option`

dnewh:
This makes the `service_id` field on the `TrackingBatch` struct an
    `Option`. This is in support of work on the batch queuer.

### DIFF
--- a/sdk/src/batch_tracking/store/diesel/models.rs
+++ b/sdk/src/batch_tracking/store/diesel/models.rs
@@ -16,6 +16,7 @@ use core::convert::TryFrom;
 use regex::Regex;
 
 use crate::batch_tracking::store::diesel::schema::*;
+use crate::batch_tracking::store::NON_SPLINTER_SERVICE_ID_DEFAULT;
 use crate::error::InternalError;
 
 use super::{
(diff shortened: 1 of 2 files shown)


sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -509,8 +514,14 @@ impl
 pub fn make_new_batch_models(batches: &[TrackingBatch]) -> Vec<NewBatchModel> {
     let mut models = Vec::new();
     for batch in batches {
+        let serv_id = if let Some(b) = batch.service_id() {
```
leebradley, line 517: The `b` isn't a great name, but you could avoid it entirely by using
```rust
let serv_id = batch.service_id().unwrap_or(NON_SPLINTER_SERVICE_ID_DEFAULT);
```

sdk/src/batch_tracking/store/diesel/models.rs
```diff
@@ -142,8 +143,12 @@ impl
             Option<SubmissionError>,
         ),
     ) -> Self {
+        let mut serv_id = Some(batch.service_id.to_string());
```
leebradley, line 146: This assignment could all be a single statement without mut

leebradley (changes requested):

sdk/src/batch_tracking/store/diesel/models.rs
dnewh: updated

sdk/src/batch_tracking/store/diesel/models.rs
dnewh: updated

leebradley (approved):

shannynalayna (approved):

PR #1416, opened 2022-05-05
Add REST API batch submission handler trait

shannynalayna:
This adds a trait for a component of the REST API that submits batches. It will take in a generic bytes representation of the batch to be submitted (or batch list) and other information pertaining to the batch to be persisted by Grid. This will be used by endpoints that accept batches, in order to process generic `TransactionPayload`s to store them as batches in Grid's DB, or to send them to an upstream server for further handling.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -127,6 +127,7 @@ experimental = [
     "proxy-client",
     "rest-api-actix-web-3",
     "rest-api-actix-web-3-run",
+    "rest-api-batch-submission-handler",
     "rest-api-resources-batch-tracking",
     "rest-api-endpoint-proxy",
     "rest-api-endpoint-record",
@@ -170,6 +171,7 @@ rest-api-actix-web-3 = [
     "url"
 ]
 rest-api-actix-web-3-run = ["rest-api-endpoint-submit"]
+rest-api-batch-submission-handler = ["rest-api-resources-batch-tracking"]
 rest-api-endpoint-agent = ["pike", "rest-api-resources-agent"]
 rest-api-endpoint-batches = ["backend", "rest-api-resources-batches"]
 rest-api-endpoint-location = ["location", "rest-api-resources-location"]
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -128,6 +128,7 @@ experimental = [
     "rest-api-actix-web-3",
     "rest-api-actix-web-3-run",
     "rest-api-batch-submission-handler",
+    "rest-api-batch-submission-handler-reqwest",
     "rest-api-resources-batch-tracking",
     "rest-api-endpoint-proxy",
     "rest-api-endpoint-record",
@@ -172,6 +173,12 @@ rest-api-actix-web-3 = [
 ]
 rest-api-actix-web-3-run = ["rest-api-endpoint-submit"]
 rest-api-batch-submission-handler = ["rest-api-resources-batch-tracking"]
+rest-api-batch-submission-handler-reqwest = [
+    "rest-api-batch-submission-handler",
+    "futures",
+    "url",
+    "reqwest",
+]
 rest-api-endpoint-agent = ["pike", "rest-api-resources-agent"]
 rest-api-endpoint-batches = ["backend", "rest-api-resources-batches"]
 rest-api-endpoint-location = ["location", "rest-api-resources-location"]
--- a/sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
+++ b/sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
@@ -1,0 +1,200 @@
+// Copyright 2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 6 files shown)


sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl ReqwestBatchSubmissionHandler {
+    pub fn new(submit_url: Url) -> Self {
+        Self { submit_url }
+    }
+}
+
+impl BatchSubmissionHandler for ReqwestBatchSubmissionHandler {
+    fn submit_batch(self, submit_request: SubmitBatchRequest) -> SubmitBatchResponse {
```
leebradley, line 36: This could probably use a test. Since it returns a future though you might need to include something like:
```toml
[dev-dependencies]
tokio = { version = "1", features = ["macros"] }
```

Idk if that version of tokio is the right one `reqwest` is resolving to, that's just off the top of my head.

And then a test like:
```rust
#[cfg(test)]
mod tests {
   use super::*;
   use tokio;
   use mockito;
   
   #[tokio::test]
   async fn requestbatchsubmissionhandler_success_correctly_returns_success() {
      let url = Url::parse(&mockito::server_url()).expect("could not parse url");
      let url = url.join("/your_path").expect("could not create url");
      
      let mock_endpoint = mock("GET", "/your_path")
            // mock a response from the server
            .with_status(200)
            .with_header("content-type", "application/json")
            .with_body("{ some json in the format your future expects }")
            .create();
            
      // ... call `submit_batch`, and await the resulting future
      // assert the future result
      
      // Ensure the endpoint was called
      mock_endpoint.assert();
   }
}
```

sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
```diff
@@ -0,0 +1,113 @@
...
+            })
+            .boxed()
+    }
+
+    fn cloned_box(&self) -> Box<dyn BatchSubmissionHandler> {
+        Box::new(self.clone())
+    }
+}
+
+impl Clone for ReqwestBatchSubmissionHandler {
```
leebradley, line 107: Pretty minor but this could just be `#[derive(Clone)]`

shannynalayna: Oh, I think I originally had to impl this because I had a `signer` field for this struct, which required a special Clone impl. But, I think now the batch builder struct will just take a signer directly from the endpoint since it's pretty easy to pass the signer around that way. Then, this submission trait isn't concerned with signing batches so the impl can be a bit more flexible.

sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
shannynalayna: Added!

sdk/src/rest_api/resources/submit/v2/payloads/batch.rs
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[derive(Serialize, Deserialize, PartialEq)]
+pub struct TrackingBatchResource {
+    pub signed_batch: BatchBytes,
+    pub service_id: Option<String>,
```
CLEckhardt, line 18: We might want to talk about if this should be a scope_id

sdk/Cargo.toml
```diff
@@ -71,6 +71,7 @@ cylinder = { version = "0.2" }
 mockito = "0.30"
 pretty_assertions = "1"
 actix-rt = "1.1.1"
+tokio = { version = "0.2", features = ["macros"] }
```
CLEckhardt, line 74: Could we call this tokio-0-2 so we can use tokio 1.0 without an alias?

sdk/src/rest_api/resources/submit/v2/payloads/batch.rs
shannynalayna: This struct will be used by endpoints for serializing/deserializing the batch objects that are passed around by the rest endpoints. So, if we do represent that as a scope ID we'd need a rest api-specific struct to mirror scope ID for (de)serialization purposes regardless. Also, the batch tracking store takes in a string for those values: https://github.com/hyperledger/grid/blob/main/sdk/src/batch_tracking/store/mod.rs#L369, linking the builder method for the object being submitted to `add_batches` (which will happen in the /batches endpoint)

CLEckhardt: Got it, thanks

sdk/src/rest_api/resources/submit/v2/handler/reqwest.rs
shannynalayna: Did add async tests, but found that the `actix_rt` dep worked here rather than tokio v1.0, so used actix rt instead

leebradley (dismissed):

CLEckhardt (approved):

leebradley (approved):

dnewh (approved):

PR #1422, opened 2022-05-19
Use `find` instead of `find_map`

shannynalayna:
The `find` Iterator method takes a closure, which returns a boolean, and
the method then returns the first object that evaluates to true as an
Option, otherwise it returns None. Previously, we were doing this
specification manually in the `find_map` method.

This issue was caught by clippy, for more information on this lint,
see: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_find_map


dnewh (approved):

CLEckhardt (approved):
Tested

peterschwarz (approved):

PR #1425, opened 2022-05-20
Update Grid to use `actix_web` 4

dnewh:
This updates grid to use `actix_web` v4. Both v3 and 4 cannot be supported in Grid due to conflicting re-exports of tokio structs/traits from the libraries. Therefore, all rest api code and tests in the Grid SDK, Griddle, and the Grid Daemon is updated to use actix_web 4.


dnewh:
This currently will not build correctly due to the old version of actix being pulled in from Splinter and causing a conflict with the Brotli decoder dependency. This update will depend on Splinter being updated to use actix 4 as well.

PR #1456, opened 2022-07-11
Jammy upgrade

rbuysse:


rbuysse:
The grid-dev:v12 image was built and published outside of CI from the commits in this PR on the same infrastructure normally used.

PR #1458, opened 2022-07-15
Add REST API module to griddle

shannynalayna:
This adds a rest api with a builder, runnable, and running instances. This allows for the rest api to be stood up and taken down more easily. The running instance of the rest api implements the `ShutdownHandle` trait from the threading::lifecycle module.  

Technically, this also updates Griddle to use actix web 4. The rest api module is behind the experimental `rest-api` feature.

### DIFF
--- a/griddle/src/rest_api/actix_web_4/api.rs
+++ b/griddle/src/rest_api/actix_web_4/api.rs
@@ -1,0 +1,44 @@
+// Copyright 2018-2022 Cargill Incorporated
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implementation for the running of the REST API, the `GriddleRestApi` handles the startup and
+//! shutdown of Griddle's REST API.
+
+use std::{future::Future, net::SocketAddr, pin::Pin, thread::JoinHandle};
+
+use actix_web::dev::ServerHandle;
+
+/// Contains information about the ports to which the REST API is bound.
+#[derive(Debug)]
+pub struct GriddleBindAddress {
+    /// The SocketAddr which defines the bound port.
+    pub addr: SocketAddr,
+
+    /// The scheme (such as http) that is running on this port.
+    pub scheme: PortScheme,
+}
+
+/// Indicates the scheme used by a port
+#[derive(Debug)]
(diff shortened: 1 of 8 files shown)


griddle/src/rest_api/actix_web_4/api.rs
```diff
@@ -0,0 +1,35 @@
...
+use actix_web::dev::Server;
+
+/// Contains information about the ports to which the REST API is bound.
+#[derive(Debug)]
+pub struct GriddleBindAddress {
+    /// The SocketAddr which defines the bound port.
+    pub addr: SocketAddr,
+
+    /// The scheme (such as http) that is running on this port.
+    pub scheme: String,
```
Caleb-Hill, line 41: Should this be an enumeration? Seems like we probably know what schemes are valid options

griddle/src/rest_api/actix_web_4/api.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022 Cargill Incorporated
+// Copyright 2018-2022 Cargill Incorporated
```
Caleb-Hill, line 1: This should probably be in the last commit?

griddle/src/rest_api/actix_web_4/api.rs
```diff
@@ -26,10 +41,149 @@ pub struct GriddleBindAddress {
...
-    server: Server,
+    server_handle: ServerHandle,
     shutdown_future: Option<Pin<Box<dyn Future<Output = ()>>>>,
 }
+
+impl GriddleRestApi {
+    pub(super) fn new(
+        bind_url: String,
+        resource_providers: Vec<Box<dyn GriddleResourceProvider>>,
+        #[cfg(feature = "proxy")] proxy_client: Box<dyn ProxyClient>,
```
Caleb-Hill, line 61: Feature gated function arguments tend to infect any structs/fns that interact with them and become hard to read. The `authorization` gated stuff is a good example in the splinter rest-api. In my opinion the codebase will be better served by a builder with optional/gated methods.

shannynalayna: Proxying is intended to be a basic functionality of Griddle, besides the batch-signing stuff, so it makes sense to me that the proxying client stays as an argument to the rest api struct. It doesn't feel like the same situation as authorization in Splinter, which someone might want more easily to turn off/on. Let me know if anyone has differing opinions on this, as I think it should stay a fn argument here. If we were to continue working on stabilizing features for Grid, this feature could be removed.

griddle/src/rest_api/actix_web_4/api.rs
shannynalayna: Updated to make this an enumeration, as we do know which schemes are valid options. The actix struct we're pulling this arg from returns a str slice, so i added a conversion method for that slice to the enum which will default to `http`.

Caleb-Hill:
lgtm

ameliabradley (approved):

ameliabradley (dismissed):

ameliabradley (approved):

CLEckhardt (approved):

